PR #24, opened 2015-01-27
Support for @Retryable in proxied beans

sparkbitpl:
This change makes it possible to find the @Retryable annotation even if
the bean originally containing it is wrapped in a proxy object. It is
useful, when the annotation is added to a method in a class and
interface proxies are used. This happens when one uses @Retryable and
@Transactional on a method implementation and proxy-target-class is set
to false.


wimdeblauwe:
Is this not a duplicate of #37 ?

PR #28, opened 2015-04-02
Fix bug BATCH-2030

shenqiao:
As reported in BATCH-2030, if BackToBackPatternClassifier has a delegate that inherits from "org.springframework.classify.Classifier", the 
context creation fails with the Exception "java.lang.IllegalStateException: More than one non-void public method detected with single argument". 
The reason is a problem in MethodInvokerUtils.getMethodInvokerForSingleArgument. Since "org.springframework.classify.Classifier" is a generic interface, class that extends from it is compiled to two public "classify" method: the bridge method and the bridged method. And in another case, this exception is also raised in IBM JVM because of the method "private java.lang.Object java.lang.Object.newInstancePrototype(java.lang.Class)".


dsyer:
Thanks. Can you confirm you've filled out the contributor's agreement (link in README: https://cla.pivotal.io/)?

dsyer:
Fixed (see b3e437b).

PR #33, opened 2015-09-03
Allow to access method metadata in retry listener

vasilievip:
I would like to have some monitoring on amount of retries happened. Looks like RetryListener is right place to start, but I do want to have at least method name which fails under retry. Updated RetryContext to have this info.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,6 +22,7 @@
 	<properties>
 		<maven.test.failure.ignore>true</maven.test.failure.ignore>
 		<spring.framework.version>4.0.4.RELEASE</spring.framework.version>
+		<maven.javadoc.failOnError>false</maven.javadoc.failOnError>
 	</properties>
 	<profiles>
 		<profile>
--- a/src/main/java/org/springframework/retry/RetryCallback.java
+++ b/src/main/java/org/springframework/retry/RetryCallback.java
@@ -16,22 +16,30 @@
 
 package org.springframework.retry;
 
+import org.aopalliance.intercept.MethodInvocation;
+
 /**
  * Callback interface for an operation that can be retried using a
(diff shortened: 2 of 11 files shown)


dsyer:
I don't like how many interfaces and APIs this change impacts. But I did something similar to support statistics anyway. So look at 8fcd85436 and notice that there is a `RetryContext.LABEL` attribute that is available in the listener, and passed down from the interceptor if you are using one.

vasilievip:
Well, I was using a bit more than just method name:

```
    @Override
    public <T, E extends Throwable> void onError(RetryContext context, RetryCallback<T, E> callback, Throwable throwable) {
        MethodInvocation methodInvocation = callback.getMethodInvocation();
        if (methodInvocation instanceof ProxyMethodInvocation) {
            methodInvocation = ((ProxyMethodInvocation) callback.getMethodInvocation()).invocableClone();
        }
        String methodName = methodInvocation.getMethod().getName();
        String className = methodInvocation.getThis().getClass().getSimpleName();
        List<Object> arguments = Arrays.asList(methodInvocation.getArguments());

        String retryMethodName = String.join(".", className, methodName);

        retryMetrics.retry(retryMethodName);

        logger.debug("Error while retry, context {}, class: {}, method {}, arguments {}", context, className, methodName, arguments, throwable);
    }
```

dsyer:
I don't think there's any more information in that than you get from `invocation.getMethod().toGenericString()`. Maybe formatted a bit differently. If you want to change it you can always set `@Retryable(label="....")`.

vasilievip:
```
        List<Object> arguments = Arrays.asList(methodInvocation.getArguments());
```
Actually, this was very valuable for debugging.

PR #36, opened 2016-01-22
Fixed incorrect package name in README

ivern:
The package name for RetryOperationsInterceptor in the XML Configuration section referenced the old Spring Batch package.


dsyer:
Fixed in 48835d7 (thanks)

PR #37, opened 2016-02-10
Find @Retryable on Target Class

### LINKED ISSUE #32: @Retryable on interface instead of implementation class
I had to put the `@Retryable` on the interface method instead of the implementation class in order to make it work, which is different how it is described in the documentation. 

```
public interface DemoService {
    @Retryable(RemoteAccessException.class)
    void service();
}
```

garyrussell:
fixes #32 

Since it's a behavior change (a previously "hidden" `Retryable` will now be found), we might need to bump to 1.2 ??


noclayto:
Any word on if these changes will be merged?

dsyer:
Merged in 51523c0

PR #39, opened 2016-02-12
support private @Recover methods

joshlong:
this makes it so that package protected / inaccessible methods are usable to `@Recover`


dsyer:
Merged in 8c24715

PR #48, opened 2016-08-11
Fix incorrect XML Configuration README.md

wetts:


dsyer:
Fixed in 48835d7 (thanks).

PR #50, opened 2016-08-30
Modified JavaDoc (SimpleRetryPolicy constructor)

privettoli:
There is no such constructor SimpleRetryPolicy(int)


dsyer:
Thanks for the patch. I think I'd prefer to simply add the constructor that is missing (it's pretty useful). See 18705bd.

PR #54, opened 2016-09-23
Set RetryState Key to null if Necessary

garyrussell:
See https://github.com/spring-projects/spring-retry/commit/9a95da2b708d5c7a0d323cfa4faf19ab10b89e81

If the key generator returns null, we need to set the key to null so the
state is not stored.


dsyer:
Applied in e6a83c0

Issue #56, opened 2016-10-04
Add SimpleRetryPolicy ctor For "Retry All Exceptions except ..."


garyrussell:
Currently, the `SimpleRetryPolicy` always creates a `BinaryExceptionClassifier` with `defaultValue=false`.

Add a ctor allowing the default to be true - retry all exceptions except those that are configured `true` in the map.

PR #57, opened 2016-10-04
GH-56: Add Retry All Except... Option

garyrussell:
Resolves: #56


dsyer:
Applied as 5a4f4c (on 1.1.x). Thanks.

PR #58, opened 2016-10-12
Support Raw Keys

garyrussell:
If the key generator is guaranteed to provide a unique key, it can be used as
the cache key directly.

Add a setter to support this option.


src/main/java/org/springframework/retry/interceptor/StatefulRetryOperationsInterceptor.java
```diff
@@ -124,6 +126,17 @@ public void setNewItemIdentifier(
 	}
 
 	/**
+	 * Set to true to use the raw key generated by the key generator. Should only be set
+	 * to true for cases where the key is guaranteed to be unique in all cases. When
+	 * false, a compound key is used, including invocation metadata.
+	 * Default: false.
+	 * @param useRawKey the useRawKey to set.
```
artembilan: `@since 1.2`, but fine with me to merge as is.

Issue #34, opened 2015-12-16
add functionality to annotation to allow a spel expressions


scottmf:
It would be very useful if the @Retryable annotation allowed an expression using spel to have an easy way to further inspect an exception in order to determine if a retry should occur.

For example:

```
@Retryable(include=HttpClientErrorException.class, retryOn="#result.getHttpStatus().value() == 401")
public void runExchange() { ... }
```

garyrussell:
You could do that relatively easily today with

    @Retryable(interceptor="myInterceptor")

Where `myInterceptor` is a `RetryOperationsInterceptor` bean with a custom retry policy that looks at the exception to decide in `canRetry()`.

scottmf:
thanks Gary.  I know the interceptor is available but in the spirit of conciseness it would be nice to have a spel expression

garyrussell:
It's probably not too hard for a single exception, but the `include` attribute is an array, and no `include` means retry for all exceptions. The expression might get unruly very quickly...

    "#this instanceof T(com.Foo) ? #this.somCondition() : #this instanceof T(com.Bar) ? #this.someOtherCondition() : false"

I am happy to discuss further, though.

scottmf:
Understood.  I would think that for more complex situations a developer would create an interceptor and for simple situations like mine it would be nice not to have to add all the extra boiler plate.  I see a lot of annotations using spel like this and situations can occur like you suggested, but for ~ 80% of the scenarios it is very clean.

I have need for the interceptor right now so that's what I am going to use.  If you strongly feel that it is the wrong thing to do then that's fine.  As a consumer of the api, who uses it a lot, it would be a nice feature to add.

garyrussell:
I just hacked together a quick PoC https://github.com/garyrussell/spring-retry/commit/f2c98b735afe5aa5ddb7075f9e58a6a73607576c

Let's see what @dsyer thinks about whether it's worth pursuing.

shollander:
Hi,
This feature would be very useful for me as well. Any word on getting the fix by @garyrussell incorporated into a release?

dsyer:
I didn't see this until now. Maybe a PR would help?

richiep33:
I am very interested in this feature as well. I have considered writing my own due to this limitation. Otherwise, spring-retry appears very useful.

garyrussell:
The PoC just evaluates an expression against the exception to decide whether it was retryable or not (which is the original request in this issue).

[...]

PR #67, opened 2016-12-01
GH-34: Resolve Placeholders in Expressions

garyrussell:
Further enhancement to #34


richiep33:
Hello Gary,

Thanks for such a quick turn around on this issue. We have tried it out and it works perfectly for allowing us to configure both the maxAttempts and backoff delay via properties in our properties file using the SPEL.

Thanks So Much,

Rich

PR #77, opened 2017-03-08
Override retry listeners for each Retryable method

maciekwiso:


src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -126,12 +126,12 @@ public void setNewItemIdentifier(NewMethodArgumentsIdentifier newMethodArguments
 
 	/**
 	 * Retry listeners to apply to all operations.
-	 * @param listeners the listeners
+	 * @param defaultListeners the listeners
 	 */
-	public void setListeners(Collection<RetryListener> listeners) {
-		ArrayList<RetryListener> retryListeners = new ArrayList<RetryListener>(listeners);
+	public void setDefaultListeners(Collection<RetryListener> defaultListeners) {
```
dsyer: This is an API change, so it's hard to justify. I would think we can leave the method name as it is (if I understand the feature correctly) and document it better? IMO this should be a list of "global" interceptors that apply to all `@Retryables``.

maciekwiso: Ok. Makes sense.

maciekwiso: Changed back the method name, not to break API

slycloud:
Hi, I'd like to overwrite default tryableListeners while using annotation `@Retryable`. But it seems that this fix is not included in release versions currently. Could I know when do you plan to release it?

dsyer:
It's in 1.2.3 (just released today - thanks for the reminder).

PR #78, opened 2017-03-17
Fix @Backoff JavaDocs - maxDelay

garyrussell:
Previously stated that the max delay is ignored if less than delay when, in fact,
the default of 30 seconds is applied.

See http://stackoverflow.com/questions/42857807/spring-retrytemplate-fires-too-early-if-no-maxdelay-is-set


dsyer:
Merged in 6c3cfe5

PR #109, opened 2018-03-26
 Calling getMaxAttempts instead of maxAttempts

plummb:
This is done to support maxAttempts to be configurable at Tenant level in a Multi Tenant Application. In this case, one can simple override getMaxAttempts and return appropriate value based on Context Tenant.


artembilan:
Merged as https://github.com/spring-projects/spring-retry/commit/da577cfeb2a804417c03f2ef47107434280f4579.

@plummb ,

sorry for so-o-o long delay.
Looking forward for more!
Now we should be able to review and merge faster 😄

PR #110, opened 2018-04-20
Enable custom RetryPolicy on Retryable and CircuitBreaker

rbraeunlich:
Add new attribute retryPolicy to CircuitBreaker and Retryable annotation to define a RetryPolicy class. This class is then picked up by the AnnotationAwareRetryOperationsInterceptor which instantiates it and passes it to the RetryTemplate.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -69,6 +69,7 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Ronny Bräunlich
  * @since 1.1
  *
  */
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@rbraeunlich Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=110)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-retry&pullRequestId=110) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@rbraeunlich Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=110)!

garyrussell:
I am not convinced this is necessary; we already have the `interceptor` attribute for more complex configuration; if we add too many options, users will be confused as to which ones to use and/or code to add mutual exclusivity would be needed.

rbraeunlich:
Hey @garyrussell 
sorry, but after four months I forgot what my actual use-case was. Just for my clarification, you're telling me I can have the same effect with an interceptor?

garyrussell:
Yes, sorry for the delay; I don't monitor this project closely.

The project provides a `RetryInterceptorBuilder` where you can build an interceptor with a fluent API to inject a rabbit template, policies etc. It gives you total control. Simply use the builder in a `@Bean` to construct an interceptor and then provide the bean name in the `interceptor` property.

rbraeunlich:
Thank you for the clarification. I guess then my PR is not necessary.

Issue #85, opened 2017-08-29
Locate closest matching recovery method given multiple retryable methods and their respective recovery methods


yashicag:
While using Spring retryable , we came across a scenario where in we have 2 methods with retryable annotation and we defined their respective recovery methods as well . 

These methods have different signature (an extra argument is present in one of the methods i.e retryable2

When retry looks for recovery method. It always considers method recover1. 

```
@Component
public class TestRetryService {

    @Retryable(maxAttempts = 3, value = NullPointerException.class)
    public void retryable1(String name) {
        System.out.println("in retryable1");
        throw new NullPointerException("from ret1");
    }

    @Retryable(maxAttempts = 3, value = NullPointerException.class)
    public void retryable2(String name, String address) {
        System.out.println("in retryable2");
        throw new NullPointerException("from ret2");
    }

    @Recover
    public void recover1(NullPointerException ne, String name ){
        System.out.println("in recover1");
    }

    @Recover
    public void recover2(NullPointerException ne, String name,String address){
        System.out.println("in recover2");
    }
}
```

Our test class ::

```
@RunWith(SpringRunner.class)
@NonTransactionalIntegrationTest
public class TestRetryServiceTest {

    @Autowired
    private TestRetryService testRetryService;

    @Test
    public void test(){

        testRetryService.retryable1("name1","add1");
    }

    @Test
    public void test2(){
        testRetryService.retryable2("name1");
    }
}
```

We see the `@Recover` interface has this statement in its documentation 
 "Subsequent arguments are populated from the argument list of the failed method in order.", however we do not see this happening. 

Also a quick glance at the snippet in `RecoverAnnotationRecoveryHandler::findClosestMatch` tells us that the algorithm looks for a closest matching recovery method based on distance between type(`Throwable`) of the recovery method(s) and the cause(i.e. value) defined in the `Retryable` (i.e. failed method).

We believe that arguments passed to recovery methods should also be considered as valid candidates to find closest matching recovery method.

crunch013:
> We believe that arguments passed to recovery methods should also be considered as valid candidates to find closest matching recovery method.

I believe that too. Or else something like Hystrix fallback method.

derylspielman:
Why not introduce a recoveryName attribute in the `@Retry(recoverName="recover1")` annotation that can then be matched in the `@Recover(name="recover1"...)` annotation? Using parameter names gets super confusing and isn't clear, which forces developers to have to do more testing of the framework just to ensure they have set it up right.

PR #125, opened 2018-08-12
Locate matching @Recover based on the number of parameters

Piszmog:
Related to https://github.com/spring-projects/spring-retry/issues/85

When locating the closest `@Recover` method and there are multiple matching `@Recover` methods, choose the `@Recover` method that matches based on the number of parameters and the types of the parameters (to avoid `java.lang.IllegalArgumentException: argument type mismatch`)

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
+++ b/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
@@ -16,16 +16,16 @@
 
 package org.springframework.retry.annotation;
 
-import java.lang.reflect.Method;
-import java.util.HashMap;
-import java.util.Map;
-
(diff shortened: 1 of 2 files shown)


derylspielman:
Why not introduce a recoveryName attribute in the `@Retry(recoverName="recover1")` annotation that can then be matched in the `@Recover(name="recover1"...)` annotation? Using parameter names gets super confusing and isn't clear, which forces developers to have to do more testing of the framework just to ensure they have set it up right.

derylspielman:
That sounds great. We would have to discuss handling scenarios such as:
* fallback return type does not match retry method return type
* handling parameter injection mismatches such as differing exception type, differing parameters
* multiple fallbacks for the same retry method differing by exception

I almost feel it should work similar to how @ControllerAdvice works where you can specify in the annotation which exceptions should be handled instead of how retry currently expects a method parameter to match the exception.

Issue #82, opened 2017-06-27
Rework Exception-wrapping


jbspeakr:
Currently, `@CircuitBreaker` always wraps failures into into an `ExhaustedRetryException`. This hides the true nature of the failure and makes it furthermore impossible to figure out whether the circuit breaker is actually open or not.

Having e.g. a `@Recover`-method just to unwrap and rethrow exceptions seems not a proper way to go... it also doesn't solve the circuit breaker status issue. 

The only currently available alternative is to implement a custom circuit breaker `RetryTemplate` which feels odd.

Please fix.

dsyer:
It's not really a `@CircuitBreaker` feature is it (the `RetryTemplate` always wraps the last failed exception)?
It has been like that since the beginning, and therefore there might be reasons and also people relying on the behaviour, so "please fix" is kind of irresponsible. If you have a concrete suggestion about how to evolve the code please feel free to comment or send patches.

PR #133, opened 2018-11-06
Add "exposeOriginalException" option on @Retryable

getupdown:
Because of issue #82, I add an option on `@Retryable` . 
It decides that if original exception ( the cause wrapped in `ExhaustedRetryException` )  will be thrown  if `ExhaustedRetryException` is triggered. 
Please review,  thanks !

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -214,6 +214,7 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		return RetryInterceptorBuilder.stateless()
 				.retryOperations(template)
 				.label(retryable.label())
+				.exposeOriginalException(retryable.exposeOriginalException())
 				.recoverer(getRecoverer(target, method))
 				.build();
 	}
@@ -239,6 +240,7 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 					.retryOperations(template)
 					.recoverer(getRecoverer(target, method))
 					.label(label)
+					.exposeOriginalException(retryable.exposeOriginalException())
 					.build();
 		}
 		RetryPolicy policy = getRetryPolicy(retryable);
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@getupdown Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=133)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-retry&pullRequestId=133) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@getupdown Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=133)!

dsyer:
Thanks for that, and for the tests. I was wondering if it might be better implemented a layer below the interceptors, so that someone using `RetryTemplate` directly could benefit. The `@Retryable` annotation might have to pick up an attribute still, or we could think about a new annotation, but let's leave that discussion for now. Do you think maybe we could add an attribute to the `RetryContext` that the `RetryTemplate` would recognize and unwrap the exception before throwing?

getupdown:
@dsyer  Thanks for your replying ! 
I've thought about your approach carefully. I found that, if `exposeOriginalException` is added into `RetryContext`, it means , `RetryPolicy` must be modified . According to the method `doExecute`, `context` is generated from `retryPolicy` and `state`. I don't think it's reasonable that the relative information should be maintained in`RetryPolicy`. 

So, how about setting a boolean value directly named  `exposeOriginalException` as a field of `RetryTemplate`  ?  

I think, the option is only for the specific implementation of `RetryOperation`, namely `RetryTemplate`.

dsyer:
Sorry for the delay. It's a tough decision. `RetryTemplate` tends to be global, but the decision to rethrow the original exception might be local. So on balance I guess you are right in the first place and we should simply support this feature in the annotation, at least to start with. Maybe I'll ask @garyrussell for an opinion before merging.

garyrussell:
From #82 @dsyer said

>It's not really a @CircuitBreaker feature is it (the RetryTemplate always wraps the last failed exception)?

Actually, the `RetryTemplate` has a boolean `throwLastExceptionOnExhausted` (false by default).

So, maybe we should add a (`default`?) setter on `RetryOperations` and set `throwLastExceptionOnExhausted` to true if the annotation has this property set; then you wouldn't need to unwrap the `ExhaustedRetryException` (because it wouldn't be wrapped in the first place).

artembilan:
Superseded by the https://github.com/spring-projects/spring-retry/commit/d3698b8cf490a6c81e8a2996c451ac7961ecb53e.

Thank you for contribution any way; looking forward for more!

Issue #89, opened 2017-09-20
ExpressionRetryPolicy should not use TemplateParserContext


artembilan:
According the documentation we can configure something like this:
```
@Retryable(exceptionExpression="#{message.contains('this can be retried')}")
public void service1() {
  ...
}
```
But since we talk about `expression` already that is redundant to ask users to wrap it into the template tokens.
Moreover this kind of requirements makes `ExpressionRetryPolicy` badly configured in the XML:
```
<bean class="org.springframework.retry.policy.ExpressionRetryPolicy">
      <constructor-arg value="#{cause.statusCode.value() == 503}" />
</bean>
```
leads to the unexpectable SpEL evaluation during bean factory initialization.
The variant like `value="cause.statusCode.value() == 503"` just make the target expression as literal.

There is indeed a workaround, but it is a bit complex for already too wordy XML config. See https://stackoverflow.com/questions/46310714/conditional-retry-advice-in-a-spring-integration-message-flow.

So, the suggestion is to do not use `TemplateParserContext` for expression parsing and just reply on the default `SpelExpressionParser.parseExpression(String expressionString)` behavior which is typical mostly for the runtime expressions in Spring Integration. 

Thanks

aldex32:
Hi @artembilan ,
I like your suggestion, but that would be backward incompatible. I would like to contribute to this issue.
Do you think adding a new property to `@Retryable`, for instance `boolean templateExpression() default true`, is a good idea? So the decision of using `TemplateParserContext` will depend on the value of `templateExpression`.
@artembilan @dsyer Let me know if you have any suggestion!

dsyer:
I don't really like the idea of adding a new attribute. Could we support the old style (while deprecating it) by just inspecting the literal before we parse it?

artembilan:
I think we indeed can support both of them meanwhile:
```
if (expressionString.contains("#{") && expressionString.contains("#{")) {
	this.expression = spelExpressionParser.parseExpression(expressionString, PARSER_CONTEXT);
}
else {
	this.expression = spelExpressionParser.parseExpression(expressionString);
}
```
Saying that *templating* does not have value in these expressions since they are really *runtime* evaluation phase against `lastThrowable` context.

PR #142, opened 2019-02-06
GH-89: Deprecate templated expression in favor of literal expression

aldex32:
resolves https://github.com/spring-projects/spring-retry/issues/89

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -27,6 +27,8 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.springframework.aop.IntroductionInterceptor;
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.BeanFactory;
@@ -35,6 +37,7 @@ import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.context.expression.BeanFactoryResolver;
 import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.core.annotation.AnnotationUtils;
+import org.springframework.expression.Expression;
 import org.springframework.expression.common.TemplateParserContext;
 import org.springframework.expression.spel.standard.SpelExpressionParser;
 import org.springframework.expression.spel.support.StandardEvaluationContext;
@@ -69,10 +72,13 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Artem Bilan
  * @author Gary Russell
+ * @author Aldo Sinanaj
  * @since 1.1
  *
  */
 public class AnnotationAwareRetryOperationsInterceptor implements IntroductionInterceptor, BeanFactoryAware {
+
+	private static final Log logger = LogFactory.getLog(ExpressionRetryPolicy.class);
 
 	private static final TemplateParserContext PARSER_CONTEXT = new TemplateParserContext();
 
@@ -256,8 +262,8 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 
 	private long getOpenTimeout(CircuitBreaker circuit) {
 		if (StringUtils.hasText(circuit.openTimeoutExpression())) {
-			Long value = PARSER.parseExpression(resolve(circuit.openTimeoutExpression()), PARSER_CONTEXT)
-					.getValue(Long.class);
+			final Expression openTimeoutExpression = getExpression(resolve(circuit.openTimeoutExpression()));
+			Long value = openTimeoutExpression.getValue(Long.class);
 			if (value != null) {
 				return value;
 			}
@@ -267,8 +273,8 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 
 	private long getResetTimeout(CircuitBreaker circuit) {
 		if (StringUtils.hasText(circuit.resetTimeoutExpression())) {
-			Long value = PARSER.parseExpression(resolve(circuit.resetTimeoutExpression()), PARSER_CONTEXT)
-					.getValue(Long.class);
+			final Expression resetTimeoutExpression = getExpression(resolve(circuit.resetTimeoutExpression()));
+			Long value = resetTimeoutExpression.getValue(Long.class);
 			if (value != null) {
 				return value;
 			}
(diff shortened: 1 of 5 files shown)


aldex32:
Hi @dsyer @artembilan ,
I fixed `exceptionExpression` to accept both templated expression (deprecated) and literal expression.
Do you guys think to do the same for the other expressions like `maxAttemptsExpression`, `delayExpression`, `maxDelayExpression` and `multiplierExpression`?
Let me know!

src/main/java/org/springframework/retry/policy/ExpressionRetryPolicy.java
```diff
@@ -114,4 +117,28 @@ public boolean canRetry(RetryContext context) {
...
 
+	/**
+	 * Check if the expression is a template
+	 * @param expression the expression string
+	 * @return true if the expression string is a template
+	 * @deprecated in favor of using literal expression
+	 */
+	@Deprecated
+	private boolean isTemplate(String expression) {
+		return expression.startsWith(PARSER_CONTEXT.getExpressionPrefix()) && expression.endsWith(PARSER_CONTEXT.getExpressionSuffix());
```
artembilan: This is not correct. The expression can be like this: `foo#{bar.value}.baz`. With meaning to evaluate a template part and concatenate it with the rest as strings.
See `ParserContext` JavaDocs.

Also that doesn't make sense to deprecate private method.

The point was to log `warn` that templating is deprecated. Nothing more we can do from the context of variable.

aldex32: fixed!

aldex32:
Hi @artembilan ,
I made other expression properties to support literal expressions.

src/main/java/org/springframework/retry/policy/ExpressionRetryPolicy.java
```diff
@@ -114,4 +119,27 @@ public boolean canRetry(RetryContext context) {
...
+
+	/**
+	 * Get expression based on the expression string.
+	 * At the moment supports both literal and template expressions. Template expressions are deprecated.
+	 * @param expression the expression string
+	 * @return literal expression or template expression
+	 */
+	private Expression getExpression(String expression) {
+		if (isTemplate(expression)) {
+			logger.warn("Template expressions are deprecated in favor of literal expressions");
```
artembilan: I would say just `Template expressions are deprecated for ExpressionRetryPolicy` .

We really can't say *literal expressions* since that is going to confuse and users just because there is really a `LiteralExpresion` to return just plain string all the time. So, we can't say that here to bring end-users in the situation when they may decide that this is trealy going to be only about simple strings.

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -416,4 +422,27 @@ private String resolve(String value) {
...
+
+	/**
+	 * Get expression based on the expression string.
+	 * At the moment supports both literal and template expressions. Template expressions are deprecated.
+	 * @param expression the expression string
+	 * @return literal expression or template expression
+	 */
+	private Expression getExpression(String expression) {
+		if (isTemplate(expression)) {
+			logger.warn("Template expressions are deprecated in favor of literal expressions");
```
artembilan: I wouldn't say this here.
As I said before, unlike `ExpressionRetryPolicy` all the other expressions are evaluated only once at start up when the annotation is parsed.

In other projects we follow a convention like support SpEL and property placeholder in the main property suffering from the fact that it can be only as a string.
For example `@RabbitListener`:
```
	/**
	 * Set to true or false, to override the default setting in the container factory.
	 * @return true to auto start, false to not auto start.
	 * @since 2.0
	 */
	String autoStartup() default "";
```
and in the `RabbitListenerAnnotationBeanPostProcessor` we do this:
```
String autoStartup = rabbitListener.autoStartup();
if (StringUtils.hasText(autoStartup)) {
	endpoint.setAutoStartup(resolveExpressionAsBoolean(autoStartup));
}
...
private Object resolveExpression(String value) {
	String resolvedValue = resolve(value);

	return this.resolver.evaluate(resolvedValue, this.expressionContext);
}

private String resolve(String value) {
	if (this.beanFactory != null && this.beanFactory instanceof ConfigurableBeanFactory) {
		return ((ConfigurableBeanFactory) this.beanFactory).resolveEmbeddedValue(value);
	}
	return value;
}
```

In other words we don't need to have here this in the `@Retryable`:
```
int maxAttempts() default 3;

String maxAttemptsExpression() default "";
```

The single `String maxAttempts() default "3";` may cover all the use-cases with simple value and expression. However again: this expression evaluated only one at start up. I call them *bean factory expresisons* since they are called against `BeanExpressionContext` and in the `StandardBeanExpressionResolver`.

I may agree that this my suggestion is going to be as a big breaking change so we may just stay with the current architecture until the next major version or  just forever...

Therefore my suggestion is to stay with templating here in the `AnnotationAwareRetryOperationsInterceptor` meanwhile supporting plain expressions. I mean there is nothing to deprecate in this class!

src/main/java/org/springframework/retry/policy/ExpressionRetryPolicy.java
garyrussell: Perhaps something more descriptive would be more helpful for novice users; something like 

    "#{...} syntax is not required for this run-time expression and is deprecated in favor of a simple expression string"

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
garyrussell: If you look at the git blame, you will see that I added the `...Expression` flavors exactly because of the breaking change.

See https://github.com/spring-projects/spring-retry/issues/34

>I went with discrete attributes to avoid breaking changes.

I don't think they can be removed until, perhaps 2.0.

But I agree; since these expressions are evaluated once-only, they should remain as-is to be consistent with normal Spring configuration conventions.

artembilan: Great! Thank you Gary!

So, @aldex32 , according my concern and Gary's feedback, please, reinstate changes in the `AnnotationAwareRetryOperationsInterceptor`.
We will remove this expression variant in the future according my explanation in the previous comment. And in that case the target expressions will remain as templated. No other choice.

Thanks for understanding and sorry for confusing if that.

src/main/java/org/springframework/retry/policy/ExpressionRetryPolicy.java
aldex32: done

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
aldex32: done

aldex32:
@artembilan @garyrussell ,
reverted the changes of `AnnotationAwareRetryOperationsInterceptor`.

aldex32:
Done!

aldex32:
Hey guys, any update? Do I need to change anything else?

garyrussell:
👍

dsyer:
LGTM

aldex32:
@artembilan seems we got the ok for this PR :-)

artembilan:
Right, we have, but the problem that I don't have permissions in this project to merge your PR.

Or @dsyer will do that himself, or he will bless my permissions here 😄 

Thanks for understanding!

dsyer:
I added @artembilan to the team. Should be good to go.

artembilan:
Merged as https://github.com/spring-projects/spring-retry/commit/2a6cd8b4472965c5445363d99aba2f27e75a7ea2 after fixing conflicts, JavaDocs, Copyright and some code style polishing in the `ExpressionRetryPolicy`.

@aldex32 ,

Thank you very much for the contribution; looking forward for more!

Issue #116, opened 2018-06-12
Allow Fluent Configuration of RetryTemplate


romeara:
A lot of spring projects allow a fluent configuration of their objects (builders/configurers) to make setup a bit easier - it would be great to add an ease-of-use option such as this. RetryInterceptorBuilder is already present, and it would nice to have something similar for RetryTemplate.

(Note: I'll try to get some time to put a PR together to propose, assuming the concept in general sounds desirable to the project, as I know the above is a little vague and it would be easier to discuss with a concrete example - I'll post snippets in here for discussion as I fiddle with it)

ashamukov:
Hello! Came up with the same idea, and found the issue:) @wpanas , if your plans change, tell, I can help

dsyer:
Closed in #153

ashamukov:
Thanks a lot!

PR #153, opened 2019-03-10
Introduction of fluent API for RetryTemplate configuration

ashamukov:
Hello! Here is an API allowing fluent configuration, like
```
RetryTemplate.newBuilder()
  .maxAttempts(10)
  .exponentialBackoff(100, 2, 10000)
  .retryOn(IOException.class)
  .traversingCauses()
  .build();

RetryTemplate.newBuilder()
  .fixedBackoff(10)
  .withinMillis(3000)
  .build();

RetryTemplate.newBuilder()
  .infiniteRetry()
  .retryOn(IOException.class)
  .uniformRandomBackoff(1000, 3000)
  .build();
```

Key class is RetryTemplateBuilder.

Implementation notes:
- SimpleRetryPolicy now is a composite policy of MaxAttemptsRetryPolicy and BinaryExceptionClassifierRetryPolicy. Having BinaryExceptionClassifierRetryPolicy separate is convenient for mixing it with another policies such as TimeoutRetryPolicy, AlwaysRetryPolicy.
- BinaryExceptionClassifierRetryPolicy and ExceptionClassifierRetryPolicy are separate classes. Using only ExceptionClassifierRetryPolicy everywhere - brings complexity, that usually is unnecessary. As well as I have not found a beautiful way to extract its common parts.
- other notes are in comments

What else can be potentially done:
- there is some logic duplication with RetryInterceptorBuilder. May be it is worth to try delegate template building part from RetryInterceptorBuilder to RetryTemplateBuilder

Questions:
- do you know sibling project's tests, that are worth to run with this changes?
- does any serialization aspects should be taken into account?
- is it worth to make 1.3 release right now, and then add the new syntax to the manual?
- other questions are in the code

Related issue: https://github.com/spring-projects/spring-retry/issues/116

### DIFF
--- a/src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
+++ b/src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
@@ -19,7 +19,9 @@ package org.springframework.retry.policy;
 import java.util.Map;
 
 import org.springframework.classify.BinaryExceptionClassifier;
+import org.springframework.retry.RetryContext;
 import org.springframework.retry.RetryPolicy;
+import org.springframework.retry.context.RetryContextSupport;
 import org.springframework.util.ClassUtils;
 
 /**
@@ -51,25 +53,23 @@ import org.springframework.util.ClassUtils;
  * @author Aleksandr Shamukov
  */
 @SuppressWarnings("serial")
-public class SimpleRetryPolicy extends CompositeRetryPolicy {
+public class SimpleRetryPolicy implements RetryPolicy {
 
 	/**
 	 * The default limit to the number of attempts for a new policy.
 	 */
 	public final static int DEFAULT_MAX_ATTEMPTS = 3;
 
-	// The reference to maxAttemptsRetryPolicy is held here to allow usage of public
-	// simpleRetryPolicy.setMaxAttempts(), while the link to exceptionClassifierRetryPolicy
-	// is held just for symmetry.
-	private final MaxAttemptsRetryPolicy maxAttemptsRetryPolicy;
-	private final BinaryExceptionClassifierRetryPolicy exceptionClassifierRetryPolicy;
+	private volatile int maxAttempts;
+
+	private BinaryExceptionClassifier retryableClassifier = new BinaryExceptionClassifier(false);
 
 	/**
 	 * Create a {@link SimpleRetryPolicy} with the default number of retry
 	 * attempts, retrying all exceptions.
 	 */
 	public SimpleRetryPolicy() {
-		this(DEFAULT_MAX_ATTEMPTS, BinaryExceptionClassifier.getDefaultClassifier());
+		this(DEFAULT_MAX_ATTEMPTS, BinaryExceptionClassifier.newDefaultClassifier());
 	}
 
 	/**
@@ -77,7 +77,7 @@ public class SimpleRetryPolicy extends CompositeRetryPolicy {
 	 * attempts, retrying all exceptions.
 	 */
 	public SimpleRetryPolicy(int maxAttempts) {
-		this(maxAttempts, BinaryExceptionClassifier.getDefaultClassifier());
+		this(maxAttempts, BinaryExceptionClassifier.newDefaultClassifier());
 	}
 
 	/**
@@ -88,7 +88,7 @@ public class SimpleRetryPolicy extends CompositeRetryPolicy {
 	 * @param retryableExceptions the map of exceptions that are retryable
 	 */
 	public SimpleRetryPolicy(int maxAttempts, Map<Class<? extends Throwable>, Boolean> retryableExceptions) {
-		this(maxAttempts, retryableExceptions, false, false);
+		this(maxAttempts, retryableExceptions, false);
 	}
 
 	/**
@@ -120,7 +120,10 @@ public class SimpleRetryPolicy extends CompositeRetryPolicy {
 	 */
 	public SimpleRetryPolicy(int maxAttempts, Map<Class<? extends Throwable>, Boolean> retryableExceptions,
 			boolean traverseCauses, boolean defaultValue) {
-		this(maxAttempts, new BinaryExceptionClassifier(retryableExceptions, defaultValue, traverseCauses));
+		super();
+		this.maxAttempts = maxAttempts;
+		this.retryableClassifier = new BinaryExceptionClassifier(retryableExceptions, defaultValue);
+		this.retryableClassifier.setTraverseCauses(traverseCauses);
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
```diff
@@ -136,72 +157,12 @@ public void setMaxAttempts(int maxAttempts) {
...
-	@Override
-	public RetryContext open(RetryContext parent) {
-		return new SimpleRetryContext(parent);
-	}
-
-	private static class SimpleRetryContext extends RetryContextSupport {
-		public SimpleRetryContext(RetryContext parent) {
-			super(parent);
-		}
-	}
```
ashamukov: Previously, there was `private static class SimpleRetryContext extends RetryContextSupport` that brings nothing new, and I've replaced it by direct usage of RetryContextSupport. Did the presence of SimpleRetryContext have any secret considerations?

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
```diff
@@ -0,0 +1,431 @@
...
+ * }</pre>
+ *
+ * <p>
+ * The builder provides the following defaults:
+ * <ul>
+ * <li> max attempts = 3 (initial + 2 retries) Inherited from {@link MaxAttemptsRetryPolicy} </li>
+ * <li> retry only on {@link Exception} and it's subclasses, without traversing causes
+ *      Inherited from {@link BinaryExceptionClassifier#getDefaultClassifier()} </li>
+ * <li> no backoff (retry immediately) </li>
+ * <li> throw an {@link org.springframework.retry.ExhaustedRetryException} on exhausted,
```
ashamukov: This it true only for stateful retry, thinking about removing this comment, or removing whole `throwLastExceptionOnExhausted` and `retryContextCache` methods from the builder API (because they are rarely used)

ashamukov: Finally, removed retryContextCache and throwLastExceptionOnExhausted from builder.

src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
dsyer: Don't know. Probably in practice not, but I think we should keep it - you can't assume that no-one ever used it, and we don't like to break existing apps.

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
dsyer: I don't like that much to be honest. Rarely used is a judgement call.

dsyer:
Seems like a good idea. I made a few comments.

src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
ashamukov: I'm also thinking how to not break something unknown, that may access this private structures.
The most safe way is to keep SimpleRetryPolicy totally unchanged. (even not to make it CompositeRetryPolicy's subclass).
The flip side is a logic duplication in SimpleRetryPolicy and MaxAttemptsRetryPolicy.
We can buy more safety by this cost.
What do you think?

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
ashamukov: Dave, let me check: do you also think that there is no sense to have `retryContextCache` and `throwLastExceptionOnExhausted` customization in builder API?

src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
ashamukov: I've added new commit where SimpleRetryPolicy structure got back, except obtaining default classifier instance from BinaryExceptionClassifier.

ashamukov:
@dsyer Dave, thanks for comments. I've reverted the structure of SimpleRetryPolicy for most safety, so, now it can be merged pretty safely.

tinesoft:
Hi,

Any idea when this great improvement will be released?
Documentation  mentions an inclusion in upcoming **v1.3.x**, but still nothing...

ashamukov:
Hi @tinesoft ! As workaround, this feature can be taken from a snapshot build, e.g.:
```  
<repositories>
    <repository>
        <id>spring-snapshots</id>
        <name>Spring Snapshots</name>
        <url>https://repo.spring.io/libs-snapshot-local</url>
        <snapshots>
            <enabled>true</enabled>
        </snapshots>
    </repository>
</repositories>

<dependencies>
    <dependency>
        <groupId>org.springframework.retry</groupId>
        <artifactId>spring-retry</artifactId>
        <version>1.3.0.BUILD-20190906.091237-9</version>
    </dependency>
        ...
```

mnisius:
Hi. I just discovered this lib yesterday and now I'm quite disappointed that I cannot use this builder pattern in my application. 

What is blocking the release for so long? Can I do something to help?

ashamukov:
Hello @mnisius ! Probably Dave @dsyer or Artem @artembilan  could answer about release circumstances. The only workaround I could suggest right now is to use snapshot build, as mentioned in the comment above (it should be safe enough, though not very beautiful of course)

mnisius:
Hi again, thx for your fast answer. Sadly I'm not allowed to use snapshot builds in my work projects. 

From a user perspective I realy think that this builder pattern is a big step forward for this library.
Just have a look at this code. It was hard to write and it is hard to read:
```java 
  @Bean
  public RetryTemplate retrier(@Value("${cos.radarlive.retries}") int retries) {
    final FixedBackOffPolicy backOffPolicy = new FixedBackOffPolicy();
    backOffPolicy.setBackOffPeriod(100);

    final ExceptionClassifierRetryPolicy retryPolicy = new ExceptionClassifierRetryPolicy();
    final SimpleRetryPolicy simpleRetryPolicy = new SimpleRetryPolicy();
    simpleRetryPolicy.setMaxAttempts(retries);
    final Map<Class<? extends Throwable>, RetryPolicy> policyMap = new HashMap<>();
    policyMap.put(WebServiceIOException.class,simpleRetryPolicy);
    retryPolicy.setPolicyMap(policyMap);

    final RetryTemplate retry = new RetryTemplate();
    retry.setBackOffPolicy(backOffPolicy);
    retry.setRetryPolicy(retryPolicy);
    return retry;
  }
```
So @dsyer und @artembilan could we please get a release anytime soon?

dsyer:
There are some outstanding PRs waiting for more input (and some that just need to be merged). We could release something and leave those for 1.4 or 2.0, but I thought the authors might want more time. If you'd like to help out you could look at those. if you just want something to use the builder is quite self contained and you can probably just copy it into your code temporarily.

mnisius:
@dsyer thx for the reply. 
This builder pattern is now wating for over 11 month for a release. So rescheduling the unfished features to an 1.4.0 release seems like a good idea for me. 

But I'll try to help if I can. Anything in particulary I can look at? 
Milestone for 1.3.0 is 100% https://github.com/spring-projects/spring-retry/milestone/9

dsyer:
From a quick scan, #141 is probably the most pressing issue. It has 2 PRs, one of which is probably just for discussion, and the other of which doesn't really solve the problem. It's probably a hard problem.

PR #168, opened 2019-03-31
Readme for fluent template configuration

ashamukov:
An example of fluent API usage is added to readme, following #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Spring for Apache Hadoop (amongst others).
 Imperative retry is also supported for explicit usage. 
 
 ## Quick Start
-#### Declarative example:
+### Declarative example
 
 ```java
 @Configuration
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,10 @@
 This project provides declarative retry support for Spring
 applications. It is used in Spring Batch, Spring Integration,
 Spring for Apache Hadoop (amongst others).
+Imperative retry is also supported for explicit usage. 
 
 ## Quick Start
-
-Example:
+#### Declarative example:
```
dsyer: This is h3 heading I think? So three hashes?

ashamukov: Initially thought about "something a bit greater than bold", but agree, this consistency is better:) Fixed.

ashamukov:
Headers have been fixed. Thanks for review!

ashamukov:
@artembilan Hi, Artem! Looks like this just need to be merged

Issue #154, opened 2019-03-16
Using ScheduledExecutorService instead of Thread.sleep


DariusX:
How about adding an implementation of Sleeper, based on ScheduledExecutorService that one could use if ThreadWaitSleeper is blocking too many threads?

ashamukov:
Hi Darius! Could you please provide an example of detailed usage scenario?
Now, I'm concerned about one thing: public API of RetryOperations is synchronous, because it returns values directly from methods.
Even if retryCallback will be executed without blocking the executing thread while backoff, caller thread (that calls execute(...)) should still blocks until whole operation completes of fails.
Consequently, to really improve whole process we should introduce asynchronous execute api first.
What do you think?

DariusX:
The use-case would be very similar to using the default sleeper. Only the implementation would not block thread. 
I was looking into another retry library, that takes an asynchronous approach. It's described here: https://www.nurkiewicz.com/2013/07/asynchronous-retry-pattern.html 
So, I was wondering if that would be a possible enhancement to Spring Retry. 
But, I did not dive into the code for Spring Retry to see if that would require a big change to the existing structure.

dsyer:
I had wondered some time ago whether `RetryTemplate` could be adapted to behave differently for a `RetryCallback` that returns an async type (like a `Future` for instance). I don't see any reason in principle why not, but the implementation is likely to get a little hairy. If anyone is interested in prototyping something that is where I would start.

ashamukov:
> I was looking into another retry library, that takes an asynchronous approach. It's described here: https://www.nurkiewicz.com/2013/07/asynchronous-retry-pattern.html
@DariusX , thanks for the link, haven't seen it before

DariusX:
I had a thought on whether the public API is currently synchronous: to the extent that the public API is _declarative_ (i.e. incorporated via annotations), it isn't really specifically synchronous. Not sure where to go with that thought, but I'm putting it out there in case is sparks an idea.

dsyer:
There is a public in `RetryTemplate` too, and `RetryCallback`, neither of which is explicitly synchronous, since they are parameterized with the return type, and that can be an async type at runtime. What is missing there is special case code inside `RetryTemplate` to deal with those return types.

`BackoffPolicy` is more likely to cause an issue. But I'm not sure it would be a complete blocker.

dsyer:
[...]

PR #176, opened 2019-05-18
An idea for async retry

ashamukov:
Hello! Here is a draft for the async retry feature with rescheduling.
The goal for now is to discuss: is it worth to implement it this way.
(javadoc and formatting are expected to be later)

The general idea is described in the README.dm diff, and also is outlined here:
async-diagram-1

**Checklist:**
- [x] solution for inter-thread retry-context propagation:
  - [x] is discussed
  - [ ] merged (+ tests)
- [ ] consensus about the value of such a feature is reached (probably, it is worth to use some full-fledged async-programming framework instead, like zio-retry for scala)
- [ ] migration of the repo to java8 is discussed and performed
- [ ] details of `java.util.concurrent.Future` support are discussed and implemented
Copy pasted from the issue:
  > Because it has no "thenApply"-like notation, exception handling and retry scheduling for actual job can not be performed by worker right after previous attempt fails, i.e. invocation of "resultFuture.get()" is required to start even the first retry of actual job. That mixes sync and async approaches in a not very beautiful way, so I propose to not specially support java.util.concurrent.Future at all. 
- [ ] javadoc and formatting are polished
- [ ] additional tests are written for:
  - [ ] stateful rescheduling
  - [ ] RejectedExecutionException on rescheduler
  - [ ] InterruptedException

**Can be done in separate PRs:**
- add ListenableFuture support
- support declarative async retry

https://github.com/spring-projects/spring-retry/issues/154

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,6 +551,48 @@ The preceding example uses a default `RetryTemplate` inside the interceptor. To 
 policies or listeners, you need only inject an instance of `RetryTemplate` into the
 interceptor.
 
+## Asynchronous retry
+### Terms
+```java
+
+CompletableFuture<HttpResponse<String>> completableFuture = retryTemplate.execute(
+                ctx -> httpClient.sendAsync(request, HttpResponse.BodyHandlers.ofString())
+        );
+```
+- __async callback__ - a callback that returns one of the supported async types (CompletableFuture, Future). Usually, async retry callback is one, that does not perform a heavy work by itself, but schedules the work to some worker and returns an instance of async type to track the progress. Failure of async callback itself usually means failure of scheduling (but not of actual work).
+Failure of async callback (of scheduling) and of actual job will both be retried on a common basis, according to configured policies.  
+
+- __job__ - a task with payload, usually heavy, which result will be available through the instance of async type, returned by async callback (and, consequently, by _execute_ method)
+
+- __rescheduling executor__ - an instance of executor, used for scheduling a new retry attempt after a delay (provided by a backoff policy). The type of executor is restricted by ScheduledExecutorService, to take advantage of its "schedule after delay" feature, which allows us to implement backoff without blocking a thread. Rescheduling executor is used for all retries except of initial scheduling retries (initial invocation of async callback).     
+
+### Initial invocation of async callback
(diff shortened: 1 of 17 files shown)


dsyer:
I rebased your work onto master. See PR: https://github.com/ashamukov/spring-retry/pull/1.

dsyer:
> About RetrySynchronizationManager
> Supported partially. There is no way to implicitly pass the context into the worker thread, because the worker executor,

True, but it's a really common pattern. The way we normally solve it in Spring projects is to provide a wrapper for the executor that has to be supplied by the user for the background state to be propagated. I expect we can do that - if the user provides the right wrapper they can use `RetrySynchronization`, otherwise not.

ashamukov:
@dsyer Hello Dave! Thanks for the rebase.

> The way we normally solve it in Spring projects is to provide a wrapper for the executor that has to be supplied by the user for the background state to be propagated.

Here is an example of solution: https://github.com/ashamukov/spring-retry/pull/2 What do you think?

artembilan:
I give that solution 👍

ashamukov:
> I give that solution 👍

Thanks, Artem!)
Are you agree, that in this situation, copy-pasting is better than extracting `DelegatingContextExecutor` to somewhere like spring-core, and parametrizing it with `RetryContext` + `RetrySynchronizationManager#register` in spring-retry, and with `SecurityContext` + `SecurityContextHolder#setContext` in spring-security?

artembilan:
Well, doesn't look like it is going to be an easy task to extract some common abstraction and we may end up with generics hell.
Although I wouldn't mind to play with that.
Do you have something to take a look already?

ashamukov:
> Do you have something to take a look already?

This is an attempt ;) https://github.com/ashamukov/spring-security/pull/1

artembilan:
You know, looks cool!

Let's raise an issue in https://github.com/spring-projects/spring-framework/issues and page @rwinch for determining what we would like to do with Spring Security after that common abstraction.

I believe there might be some improvements or changes in your solution, but at least my colleagues will know our position about such a propagation abstraction from one thread to another! 😄

ashamukov:
Thanks, Artem!
> Let's raise an issue

The issue: https://github.com/spring-projects/spring-framework/issues/23876

dsyer:
I rebased and did some tidying up: https://github.com/dsyer/spring-retry/tree/ashamukov-async-retry. We can start from there if anyone is interested in this. It works for `CompletableFuture` and `Future` (as least as much as is tested).

artembilan:

PR #182, opened 2019-10-03
Fix #126 clearer documentation, enforce maxInterval value

### LINKED ISSUE #126: ExponentialRandomBackOffPolicy counts incorrect sleep intervals
Hey there, 

Spec says:
Example: initialInterval = 50 multiplier = 2.0 maxInterval = 3000 numRetries = 5 ExponentialBackOffPolicy yields: [50, 100, 200, 400, 800] ExponentialRandomBackOffPolicy may yield [50, 100, 100, 100, 600] or [50, 100, 150, 400, 800]

However if I look to implementation
```
public synchronized long getSleepAndIncrement() {
			long next = super.getSleepAndIncrement();
			next = (long) (next * (1 + r.nextFloat() * (getMultiplier() - 1)));
			return next;
		}
```

It means result will be never less than super method result. So either spec is wrong or implementation. Also tested several times.  Am I missing something?

Thanks and Cheers

abracadv8:
Fix #126 -- Clearer documentation on what ranges are possible in the example, check to enforce maxInterval so the random multiplication step cannot exceed it.

Pivotal CLI signed for a previous Spring project.  Contribution on behalf of Comcast Corporation.

### DIFF
--- a/src/main/java/org/springframework/retry/backoff/ExponentialRandomBackOffPolicy.java
+++ b/src/main/java/org/springframework/retry/backoff/ExponentialRandomBackOffPolicy.java
@@ -37,10 +37,12 @@ import java.util.Random;
  * {@link ExponentialBackOffPolicy} yields: [50, 100, 200, 400, 800]
  *
  * {@link ExponentialRandomBackOffPolicy} may yield [76, 151, 304, 580, 901] or [53, 190,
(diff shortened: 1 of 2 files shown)


abracadv8:
Closing to clean up history a bit, and creating a new pull.

PR #187, opened 2019-11-08
Allow recover by name in `Retryable` annotation.

kitaisreal:
> Why not introduce a recoveryName attribute in the `@Retry(recoverName="recover1")` annotation that can then be matched in the `@Recover(name="recover1"...)` annotation? Using parameter names gets super confusing and isn't clear, which forces developers to have to do more testing of the framework just to ensure they have set it up right.
https://github.com/spring-projects/spring-retry/issues/85

**Examples of usage**

``` java
protected static class RecoverByRetryableName {

	@Retryable(recoverName = "barRecover")
	public int foo(String name) {
		return 0;
	}

	@Recover
	public int fooRecover(Throwable throwable, String name) {
		return 1;
	}

	@Recover
	public int barRecover(Throwable throwable, String name) {
		return 2;
	}
}
```

**Current implementation steps**
1.  Added recover name in Retryable annotation.
3. Updated tests.

**To be done before merge**
1. Update Readme
1. Update javadoc

### DIFF
--- a/README.md
+++ b/README.md
@@ -481,25 +481,24 @@ The following example shows how to do so:
 
 ```java
 @Service
-class Service {
-
+class Service { 
     @Retryable(recover = "service1Recover", value = RemoteAccessException.class)
-	public void service1(String str1, String str2) {
+    public void service1(String str1, String str2) {
         // ... do something
     }
 
     @Retryable(recover = "service2Recover", value = RemoteAccessException.class)
-	public void service2(String str1, String str2) {
+    public void service2(String str1, String str2) {
         // ... do something
     }
 
     @Recover
-	public void service1Recover(RemoteAccessException e, String str1, String str2) {
+    public void service1Recover(RemoteAccessException e, String str1, String str2) {
         // ... error handling making use of original args if required
     }
 
     @Recover
-	public void service2Recover(RemoteAccessException e, String str1, String str2) {
+    public void service2Recover(RemoteAccessException e, String str1, String str2) {
         // ... error handling making use of original args if required
     }
 }
--- a/README.md
+++ b/README.md
@@ -486,13 +486,11 @@ class Service {
     @Retryable(recover = "service1Recover", value = RemoteAccessException.class)
 	public void service1(String str1, String str2) {
         // ... do something
-		throw new RemoteAccessException("");
     }
 
     @Retryable(recover = "service2Recover", value = RemoteAccessException.class)
 	public void service2(String str1, String str2) {
         // ... do something
-		throw new RemoteAccessException("");
     }
 
     @Recover
--- a/README.md
+++ b/README.md
@@ -476,6 +476,37 @@ class Service {
 }
 ```
 
+To resolve conflicts between multiple methods that can be picked for recovery. You can explicitly specify recovery method name. 
+The following example shows how to do so:
+
+```java
+@Service
+class Service {
+
+    @Retryable(recover = "service1Recover", value = RemoteAccessException.class)
+	public void service1(String str1, String str2) {
+        // ... do something
+		throw new RemoteAccessException("");
+    }
+
+    @Retryable(recover = "service2Recover", value = RemoteAccessException.class)
+	public void service2(String str1, String str2) {
+        // ... do something
+		throw new RemoteAccessException("");
+    }
+
+    @Recover
+	public void service1Recover(RemoteAccessException e, String str1, String str2) {
+        // ... error handling making use of original args if required
+    }
+
+    @Recover
+	public void service2Recover(RemoteAccessException e, String str1, String str2) {
+        // ... error handling making use of original args if required
+    }
+}
+```
+
 Version 1.2 introduces the ability to use expressions for certain properties. The
 following example show how to use expressions this way:
 
--- a/src/main/java/org/springframework/retry/annotation/Recover.java
+++ b/src/main/java/org/springframework/retry/annotation/Recover.java
@@ -42,11 +42,4 @@ import org.springframework.context.annotation.Import;
 @Documented
 public @interface Recover {
 
-	/**
-	 * Name of this function can be used to provide best match for recoverName in
-	 * Retryable. {@link Retryable#recoverName()}
-	 * @return name of this recover function
-	 */
-	String name() default "";
-
 }
--- a/src/main/java/org/springframework/retry/annotation/Recover.java
+++ b/src/main/java/org/springframework/retry/annotation/Recover.java
@@ -42,4 +42,11 @@ import org.springframework.context.annotation.Import;
 @Documented
 public @interface Recover {
 
+	/**
+	 * Name of this function can be used to provide best match for recoverName in
+	 * Retryable. {@link Retryable#recoverName()}
+	 * @return name of this recover function
+	 */
+	String name() default "";
+
 }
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@kitaisreal Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=187)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-retry&pullRequestId=187) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@kitaisreal Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=187)!

src/main/java/org/springframework/retry/annotation/Recover.java
```diff
@@ -42,4 +42,11 @@
 @Documented
 public @interface Recover {
 
+	/**
+	 * Name of this function can be used to provide best match for recoverName in
+	 * Retryable. {@link Retryable#recoverName()}
+	 * @return name of this recover function
+	 */
+	String name() default "";
```
artembilan: Why do we need this?

Why just method name in the `@Retryable` attribute is not enough?

src/main/java/org/springframework/retry/annotation/Retryable.java
```diff
@@ -36,6 +36,12 @@
 @Documented
 public @interface Retryable {
 
+	/**
+	 * Name of function to use for recover.
+	 * @return the name of retry function to use
+	 */
+	String recoverName() default "";
```
artembilan: I think `recover` is fully enough to indicate that we would like to have a strict method reference

src/main/java/org/springframework/retry/annotation/Recover.java
kitaisreal: I agree, better to have one way of providing explicit recovery method name in `@Retryable` without confusing client with additional name in `@Recover`. So I removed additional parameter from `@Recover`.

src/main/java/org/springframework/retry/annotation/Retryable.java
kitaisreal: Agree.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -81,25 +83,42 @@ public T recover(Object[] args, Throwable cause) {
...
-		int min = Integer.MAX_VALUE;
 		Method result = null;
-		for (Method method : this.methods.keySet()) {
-			SimpleMetadata meta = this.methods.get(method);
-			Class<? extends Throwable> type = meta.getType();
-			if (type == null) {
-				type = Throwable.class;
+
+		boolean shouldUseRecoverMethodName = !this.recoverMethodName.equals("");
+		if (shouldUseRecoverMethodName) {
```
artembilan: We don't need this variable. The `StringUtils.isEmpty()` is or friend in this case.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -81,25 +83,42 @@ public T recover(Object[] args, Throwable cause) {
...
+			for (Method method : this.methods.keySet()) {
```
artembilan: The map iteration must be done over `entrySet`.
This way we won't have an overhead requesting each entry value with the next `get`.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -81,25 +83,42 @@ public T recover(Object[] args, Throwable cause) {
...
 			}
-			if (type.isAssignableFrom(cause)) {
-				int distance = calculateDistance(cause, type);
-				if (distance < min) {
-					min = distance;
-					result = method;
+		}
+		else {
+			int min = Integer.MAX_VALUE;
+			for (Method method : this.methods.keySet()) {
```
artembilan: Same here about iteration over an `entrySet()`

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -139,6 +158,8 @@ private boolean compareParameters(Object[] args, int argCount, Class<?>[] parame
 	private void init(Object target, Method method) {
 		final Map<Class<? extends Throwable>, Method> types = new HashMap<Class<? extends Throwable>, Method>();
 		final Method failingMethod = method;
+		Retryable retryable = method.getAnnotation(Retryable.class);
+		this.recoverMethodName = retryable != null ? retryable.recoverName() : "";
```
artembilan: I think `null` by default is OK, when we start to use a `StringUtils.isEmpty()` mentioned above

src/main/java/org/springframework/retry/annotation/Retryable.java
artembilan: Well, but you still haven't renamed it just into a `recover()`.
And why do you call it `funciton` in the JavaDocs?
Isn't the story just about an annotated method in the same class where we have this `@Retryable`?
Why just don't explain that in this JavaDocs?

Also I think the `function` may confuse when we use a Spring Cloud Function. So, one may think this is a name of the function bean. Therefore I think we need to be more explicit and say that we want to have here a method name.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
kitaisreal: Agree. Fixed.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
kitaisreal: Agree. Fixed.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
kitaisreal: Fixed.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
kitaisreal: Agree. Fixed.

src/main/java/org/springframework/retry/annotation/Retryable.java
kitaisreal: Sure. Updated documentation for this field.

kitaisreal:
Yes, sure, updated

README.md
```diff
@@ -476,6 +476,37 @@ class Service {
...
 
+To resolve conflicts between multiple methods that can be picked for recovery. You can explicitly specify recovery method name. 
+The following example shows how to do so:
+
+```java
+@Service
+class Service {
+
+    @Retryable(recover = "service1Recover", value = RemoteAccessException.class)
+	public void service1(String str1, String str2) {
```
artembilan: Please, consider to not use *tabs* as indents in docs - just *4 spaces* instead.
Otherwise you see how code is not aligned properly

src/main/java/org/springframework/retry/annotation/Retryable.java
```diff
@@ -38,7 +38,7 @@
 public @interface Retryable {
 
 	/**
-	 * Name of method in this class to use for recover.
+	 * Name of method in this class to use for recover. Method had to be marked with {@link Recover} annotation.
```
artembilan: Please, move this new sentence on a new line.

README.md
kitaisreal: Sure. Fixed.

artembilan:
@kitaisreal ,

thank you for the contribution; looking forward for more!

PR #124, opened 2018-08-10
Find @Recover/CB on target class with JDK proxy

garyrussell:
Similar to https://github.com/spring-projects/spring-retry/pull/37

`@Recover` is only detected on interfaces for JDK proxies whereas
`@Retryable` is found on the target class if not on the interface.

PR #199, opened 2020-04-07
Support @Recovery method that is not defined in interface

### LINKED ISSUE #197: Recovery method not found for method in class that implements an interface
I have a component class that implements an interface. The interface is being autowired into my calling class, and in my implementation class I have annotated a method from the interface with `@Retryable` and added a public recovery method. However, when the retry fails, it fails to find the recovery method.

Through some debugging, I have observed that `ReflectionUtils#doWithMethods(Class, MethodCallback)` is used to iterate through the methods of the failing method's class to find the recovery method. However, in my case, `failingMethod.getDeclaringClass()` is returning my interface instead of my implementation class, and because the interface doesn't contain my recovery method, the recovery method is never found. I noticed that the object of my implementation class is passed to the `RecoverAnnotationRecoveryHandler#init(Object, Method)` method in the target parameter, but never used.

```
	private void init(Object target, Method method) {
		final Map<Class<? extends Throwable>, Method> types = new HashMap<Class<? extends Throwable>, Method>();
		final Method failingMethod = method;
		ReflectionUtils.doWithMethods(failingMethod.getDeclaringClass(),
```

Is there a reason that `failingMethod.getDeclaringClass()` is being used instead of `target.getClass()`? Thank you for looking at this issue. Let me know if I should open a PR to attempt to fix this.

Currently using Spring Retry 1.2.5.RELEASE with Spring Boot 2.2.5.RELEASE.

engineal:
Fixes #197. Even with the changes from #124, recovery methods that are not defined in the interface are not found. I've tested my suggested change, replacing `failingMethod.getDeclaringClass()` with `target.getClass()`. No unit tests failed with this change 🤞.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
+++ b/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
@@ -166,7 +166,7 @@ public class RecoverAnnotationRecoveryHandler<T> implements MethodInvocationReco
 		if (retryable != null) {
 			this.recoverMethodName = retryable.recover();
 		}
-		ReflectionUtils.doWithMethods(failingMethod.getDeclaringClass(), new MethodCallback() {
+		ReflectionUtils.doWithMethods(target.getClass(), new MethodCallback() {
 			@Override
 			public void doWith(Method method) throws IllegalArgumentException, IllegalAccessException {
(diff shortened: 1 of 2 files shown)


dsyer:
Thanks. That looks useful. @garyrussell is there any downside? Any way for the target class to not have all the methods we need?

Issue #205, opened 2020-06-03
NPE when circuit tries to open @CircuitBreaker. 


Casvad:
Hi im getting this error on production with this configuration 
`@CircuitBreaker(resetTimeout = 15000, openTimeout = 1000, maxAttempts = 2)`

`Caused by: java.lang.NullPointerException at org.springframework.retry.policy.CircuitBreakerRetryPolicy$CircuitBreakerRetryContext.isOpen(CircuitBreakerRetryPolicy.java:152) at org.springframework.retry.policy.CircuitBreakerRetryPolicy.canRetry(CircuitBreakerRetryPolicy.java:77) at org.springframework.retry.support.RetryTemplate.canRetry(RetryTemplate.java:375) at org.springframework.retry.support.RetryTemplate.doExecute(RetryTemplate.java:278) at org.springframework.retry.support.RetryTemplate.execute(RetryTemplate.java:211) at org.springframework.retry.interceptor.StatefulRetryOperationsInterceptor.invoke(StatefulRetryOperationsInterceptor.java:173) at org.springframework.retry.annotation.AnnotationAwareRetryOperationsInterceptor.invoke(AnnotationAwareRetryOperationsInterceptor.java:153) `

this seems to happened when the this attribute is not set 
<img width="690" alt="image" src="https://user-images.githubusercontent.com/17624447/83663889-e4c0c880-a58e-11ea-82b2-cfbad18ec538.png">

PR #207, opened 2020-06-03
fix issue (#205) null pointer exception when circuit break attribute is not setted (CircuitBreakerRetryPolicy)

oscarcitoz:


pivotal-issuemaster:
@oscarcitoz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-retry&pullRequestId=207)!

artembilan:
Please, mention an issue https://github.com/spring-projects/spring-retry/issues/205 and consider to add some test to cover this problem.

Thanks

oscarcitoz:
> Please, mention an issue #205 and consider to add some test to cover this problem.
> 
> Thanks

@artembilan already add test and i mention issue

src/test/java/org/springframework/retry/policy/MockNeverRetryPolicy.java
```diff
@@ -0,0 +1,9 @@
+package org.springframework.retry.policy;
+
+import org.springframework.retry.RetryContext;
+
+public class MockNeverRetryPolicy extends NeverRetryPolicy {
```
artembilan, line 5: I'm not sure why we need this class as top-level one: probably really better to have it as inner one, like that `MockRetryCallback` in the `CircuitBreakerRetryTemplateTests`. But still let's reconsider if we just can deal with plain `NeverRetryPolicy`!

oscarcitoz: Hi @artembilan, I did the mock for this class because I wanted to show you fix the bug (NPE), then in this case every executions in the test will enter in that point. no matter what class you use, the important is show the fix.  if you want, can you run the test without my fix for you understand the problem and the solution (and the test).

oscarcitoz: likewise I will change the class with your suggestion

artembilan:
@oscarcitoz ,

thank you very much for the contribution; looking forward for more!

Issue #244, opened 2021-06-04
@Recover methods not intercepted by Spring AOP advice


kriegaex:
I just answered a question on StackOverflow, please read for more details and my analysis:
https://stackoverflow.com/a/67831919/1082681

The original poster of the question also published this reproducer which helped me find the root cause:
https://github.com/Nikhilgupta1891/RetryRecover

In a nutshell, Spring AOP cannot intercept the execution of a `@Recover` method because of `invocation.getThis()` instead of `invocation.getProxy()` here:

https://github.com/spring-projects/spring-retry/blob/e637911a8af4bbef372824cb9e13e3b8fa67426b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java#L159-L168

If there is no compelling reason to use the target object instead of its corresponding proxy (if any exists), please change it accordingly.

garyrussell:
Changing it to `getProxy()` caused other problems.

I did come up with a solution, but it's a bit of a hack.

https://github.com/spring-projects/spring-retry/pull/245

kriegaex:
I locally compiled and installed `1.3.2-SNAPSHOT` from the latest master and tested it against the sample repository provided by the other user. The aspect now kicks in as expected. Thanks for the quick fix on behalf of him.

PR #245, opened 2021-06-04
GH-244: Invoke Proxy for @Recover Method

garyrussell:
Resolves https://github.com/spring-projects/spring-retry/issues/244

The `RecoverAnnotationRecoveryHandler` only has access to the undecorated
raw bean.

Use the `RetryContext` to make the proxy (if CGLIB) available to the recoverer.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
+++ b/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
@@ -22,10 +22,13 @@ import java.lang.reflect.Type;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.springframework.aop.support.AopUtils;
 import org.springframework.classify.SubclassClassifier;
 import org.springframework.core.annotation.AnnotationUtils;
 import org.springframework.retry.ExhaustedRetryException;
+import org.springframework.retry.RetryContext;
 import org.springframework.retry.interceptor.MethodInvocationRecoverer;
+import org.springframework.retry.support.RetrySynchronizationManager;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.ReflectionUtils.MethodCallback;
 import org.springframework.util.StringUtils;
@@ -76,8 +79,19 @@ public class RecoverAnnotationRecoveryHandler<T> implements MethodInvocationReco
 		boolean methodAccessible = method.isAccessible();
 		try {
 			ReflectionUtils.makeAccessible(method);
+			RetryContext context = RetrySynchronizationManager.getContext();
+			Object proxy = null;
+			if (context != null) {
+				proxy = context.getAttribute("___proxy___");
+				if (AopUtils.isJdkDynamicProxy(proxy)) {
+					proxy = null;
+				}
+			}
+			if (proxy == null) {
+				proxy = this.target;
+			}
 			@SuppressWarnings("unchecked")
-			T result = (T) ReflectionUtils.invokeMethod(method, this.target, argsToUse);
+			T result = (T) ReflectionUtils.invokeMethod(method, proxy, argsToUse);
 			return result;
 		}
 		finally {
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -76,8 +79,19 @@ public T recover(Object[] args, Throwable cause) {
...
+			RetryContext context = RetrySynchronizationManager.getContext();
+			Object proxy = null;
+			if (context != null) {
+				proxy = context.getAttribute("___proxy___");
+				if (AopUtils.isJdkDynamicProxy(proxy)) {
+					proxy = null;
+				}
+			}
+			if (proxy == null) {
+				proxy = this.target;
```
artembilan, line 91: What is the problem if we always propagate a proxy to this instance not just target?

src/main/java/org/springframework/retry/interceptor/RetryOperationsInterceptor.java
```diff
@@ -111,9 +115,15 @@ public Object doWithRetry(RetryContext context) throws Exception {
...
-		if (recoverer != null) {
-			ItemRecovererCallback recoveryCallback = new ItemRecovererCallback(invocation.getArguments(), recoverer);
-			return this.retryOperations.execute(retryCallback, recoveryCallback);
+		if (this.recoverer != null) {
+			ItemRecovererCallback recoveryCallback = new ItemRecovererCallback(invocation.getArguments(),
+					this.recoverer);
+			Object recovered = this.retryOperations.execute(retryCallback, recoveryCallback);
+			RetryContext context = RetrySynchronizationManager.getContext();
+			if (context != null) {
+				context.removeAttribute("__proxy__");
```
artembilan, line 124: Shouldn't this go to the `finally` then?

artembilan (changes requested):

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
garyrussell: 1. There are some tests that call this directly, so no proxy is added to the context.
2. If the proxy is a JDK proxy we can't invoke the recoverer because the method is the instance method, not the interface method.

```
java.lang.IllegalArgumentException: object is not an instance of declaring class
```

kriegaex: 1. I am not sure if designing a change not to break existing tests is such a good idea. Maybe you should updatte the tests instead.
2. Even if the proxy is a JDK proxy, a proxied interface method could point to a method in the target instance which is annotated by `@Recover`. Can you explain what exactly does not work there? Is it just difficult to discover the target method or is there some other basic problem which I simply don't understand, being an AspectJ expert but a relative SpringAOP layman?

kriegaex:
Independent of my questions above: Thanks for the lightning-fast reaction and quick fix. This surely helps a bunch of people. :-)

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
garyrussell: 1. Good point; I was just being lazy to get a quick fix out.
2. Yes, we can add code to find the corresponding method on the proxy (if it exists). The problem is it won't solve every use case because the proxied interfaces may not expose the recover method. So we'd still need to handle the case where we call the recover method directly on the target instead of the proxy.

kriegaex: 2. Yes, I know. In this case...
```java
public interface Foo {
  String bar();
```
```java
@Component public class FooImpl {
  @Override public String bar() { return "bar"; }
  @Recover public int notFromInterface() { return 42; }
}
```
... of course it will not work, because `notFromInterface` is not exposed by a JDK proxy.

But this would be detectable:

```java
@Component public class FooImpl {
  @Override @Recover public String bar() { return "bar"; }
}
```

artembilan (changes requested):
LGTM, but how about to cover this feature with some unit tests?

Thanks

Issue #214, opened 2020-08-31
Spring-Retry @Retryable not working with Spring-data Repositories


jopuneet:
I was trying @Retryable with Spring-data and testing it out using Mockito @MockBean but the retry for the repository is not working and it was always called only once.

```
@Repository
public interface TestRepository extends CrudRepository<Test, String> {

  @Retryable(value = { Exception.class }, maxAttempts = 4, backoff = @Backoff(delay = 2000))
  List<Test> findAll();
}
```
```
when(testRepository.findAll())
            .thenThrow(new RecoverableDataAccessException("test RecoverableDataAccessException 1"))
            .thenThrow(new RecoverableDataAccessException("test RecoverableDataAccessException 2"))
            .thenReturn(Lists.newArrayList(test));
verify(testRepository, times(3)).findAll(any());
```
```
org.mockito.exceptions.verification.TooFewActualInvocations: 
testRepository bean.findAll(<any string>);
Wanted 3 times:
-> at com.*.findAllTest(TestIT.java:10)
But was 1 time:
```

Referring to this [Stackoverflow answer](https://stackoverflow.com/questions/42730104/retryable-doesnt-work-in-spring-data-repository-interface)

Spring-boot: 2.3.0.RELEASE
Spring-Retry: 1.2.5.RELEASE
Spring-data: 2.3.0.RELEASE

garyrussell:
If you are mocking the `TestRepository` then it's a mock instead of a Spring Bean so, of course, retry won't work; Mockito doesn't know anything about Spring.

Puneetjoshi24:
@garyrussell I am using `@MockBean` to mock the `TestRepository` not `@Mock` and `@MockBean` provide spring context with the mock. As referenced here [Baeldung MockBean](https://www.baeldung.com/java-spring-mockito-mock-mockbean#:~:text=We%20can%20use%20the%20%40MockBean,type%20in%20the%20application%20context.&text=When%20we%20use%20the%20annotation,be%20injected%20into%20the%20field.) 

If spring retry will not work with Mock here, what is the way to test Spring-retry over Spring-Jpa? We already have a question on [Stackoverflow](https://stackoverflow.com/questions/42730104/retryable-doesnt-work-in-spring-data-repository-interface) .

Puneetjoshi24:
[...]

PR #246, opened 2021-06-08
GH-214: Retryable on JPA Repos with proxyTargetClass [DO NOT MERGE]

garyrussell:
Resolves https://github.com/spring-projects/spring-retry/issues/214

When `proxyTargetClass` is `true` (Boot's default), the JPA interfaces
are not copied to the outer retry proxy created by auto-proxy.

Add a BPP to fix up the proxy, when needed.

Also, the pointcut filter must match on the `SimpleJpaRepository`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,13 @@ line  to your `build.gradle` file:
     runtime('org.aspectj:aspectjweaver:1.8.13')
 ```
 
+### Using Declarative Retry on Spring JPA Repository Interfaces
+
+When using Spring Spring Boot, to enable support for adding `@Retryable` to JPA Repository Interface methods, you can add
+a `JPARepositoryRetryBeanPostProcessor` to the application context, or set the [spring.aop.proxy-target-class](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.core.spring.aop.proxy-target-class) property to `false`.
+When this property is true, it prevents the auto-proxy mechanism from copying the repository interfaces to the retry proxy.
+Adding the bean post processor causes a new proxy to be created, merging the interfaces and advisors on the JPA and retry proxies.
+
 ### XML Configuration
 
 The following example of declarative iteration uses Spring AOP to repeat a service call to
--- a/pom.xml
+++ b/pom.xml
@@ -276,6 +276,12 @@
 			<version>1.2.17</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-jpa</artifactId>
+			<version>2.5.1</version>
+			<optional>true</optional>
+		</dependency>
 	</dependencies>
 
 	<reporting>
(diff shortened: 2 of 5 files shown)


garyrussell:
cc/ @mp911de

README.md
```diff
@@ -596,6 +596,13 @@ line  to your `build.gradle` file:
     runtime('org.aspectj:aspectjweaver:1.8.13')
 ```
 
+### Using Declarative Retry on Spring JPA Repository Interfaces
```
mp911de, line 599: I'd suggest to generally allow Spring Data repositories.

pom.xml
```diff
@@ -276,6 +276,12 @@
 			<version>1.2.17</version>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.data</groupId>
```
mp911de, line 280: I'd suggest using Spring Data Commons to enable retry functionality for all Spring Data modules.

src/main/java/org/springframework/retry/annotation/JPARepositoryRetryBeanPostProcessor.java
```diff
@@ -0,0 +1,79 @@
...
+public class JPARepositoryRetryBeanPostProcessor implements BeanPostProcessor, Ordered {
+
+	@Override
+	public int getOrder() {
+		return Integer.MAX_VALUE;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
+		if (bean instanceof Retryable && bean instanceof Advised) {
```
mp911de, line 44: You could generally check if the bean implements `org.springframework.data.repository.Repository` (as marker for Spring Data repositories) to make this work for all Spring Data modules. Not sure why JpaRepositoryImplementation` needs to be excluded here.

mp911de (commented):
Let me know if we want to dig into a discussion to make this work for all Spring Data repository implementations. Would be probably easier than keeping code for all kinds of Spring Data modules around.

src/main/java/org/springframework/retry/annotation/JPARepositoryRetryBeanPostProcessor.java
garyrussell: Thanks Mark; without removing that interface, I get

```
Caused by: java.lang.IllegalStateException: No MethodInvocation found: Check that an AOP invocation is in progress, and that the CrudMethodMetadataPopulatingMethodInterceptor is upfront in the interceptor chain.
	at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$CrudMethodMetadataPopulatingMethodInterceptor.currentInvocation(CrudMethodMetadataPostProcessor.java:129) ~[spring-data-jpa-2.5.1.jar:2.5.1]
	at org.springframework.data.jpa.repository.support.CrudMethodMetadataPostProcessor$ThreadBoundTargetSource.getTarget(CrudMethodMetadataPostProcessor.java:322) ~[spring-data-jpa-2.5.1.jar:2.5.1]
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:195) ~[spring-aop-5.3.7.jar:5.3.7]
```

I am going to close this PR for now, and open an issue against spring-aop, to see if we can get the root cause fixed.

garyrussell: https://github.com/spring-projects/spring-framework/issues/27044

Let's see what the framework guys say before doing any more.

PR #247, opened 2021-06-15
Allow primitive type for @Recover method parameter

leeyohan93:
I want to use the primitive type when I specify the Recover method.

But I could only use the wrapped type like below and I simply fixed this part.

I simply fixed this part, so I think you can fix it a better.

Thank you.

```java
@Retryable(recover = "barRecover")
public int foo(int number);

@Recover
public int fooRecover(Throwable throwable, Integer number);

@Recover
public int barRecover(Throwable throwable, Integer number);
```


artembilan:
Merged as https://github.com/spring-projects/spring-retry/commit/4b298b3b07f3402637945c7d6437779e0bdeb313.

@leeyohan93 , 

thank you for the contribution; looking forward for more!

Issue #256, opened 2022-01-13
RetryConfiguration is not eligible for getting processed by all BeanPostProcessors...


markusheiden:
I get always this info message from the BeanPostProcessorChecker during startup:

```
BeanPostProcessorChecker : Bean 'org.springframework.retry.annotation.RetryConfiguration' of type [org.springframework.retry.annotation.RetryConfiguration] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
```

This is most probably due to the injections in RetryConfiguration. BeanPostProcessorChecker would suppress this warning if RetryConfiguration would be annotated with `@Role(ROLE_INFRASTRUCTURE)`.

PR #257, opened 2022-01-14
Mark RetryConfiguration as infrastructure bean

markusheiden:
This suppresses the info message "Bean 'org.springframework.retry.annotation.RetryConfiguration' of type [org.springframework.retry.annotation.RetryConfiguration] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)" during the Spring context start.

Relates to issue #256

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
+++ b/src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014-2021 the original author or authors.
+ * Copyright 2014-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
```diff
@@ -51,6 +52,8 @@
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.ReflectionUtils.MethodCallback;
 
+import static org.springframework.beans.factory.config.BeanDefinition.ROLE_INFRASTRUCTURE;
```
artembilan, line 55: We prefer to avoid `import static` in production code. 
While it might fix the immediate warning, I wonder if it is still legit to keep the logic this `InitializingBean` way...
May be better to revise it in favor of `SmartInitializingSingleton`, so there won't be dependencies on other beans during this `Advisor` introspection.
I don't tell that we need to do this right now since it might be slightly a breaking change even if this class is not intended to be used somehow in the target project.

Please, update the copyright header to he current year and add your name to the `@author` list.
Then we can merge this as is and revise the possibilities later on.
Thanks

artembilan (changes requested):

markusheiden:
I changed what you requested.

src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
markusheiden: I changed what you requested.

artembilan:
Thank you, @markusheiden , for contribution; looking forward for more!

Issue #264, opened 2022-01-26
@Retryable - Support Rethrow Instead of ExhaustedRetryException


garyrussell:
See https://github.com/spring-projects/spring-retry/discussions/263#discussioncomment-2051710

garyrussell:
I propose the following as a straw man:

Add 

`@AliasFor("include") Class<? extends Throwable>[] retryFor()`

`@AliasFor("exclude") Class<? extends Throwable>[] noRetryFor()`

`Class<? extends Throwable>[] noRetryNoRecoveryFor()`

All others not specified are not retriable, but go straight to recovery.

No `rethrow` attribute.

artembilan:
Sounds like a plan.
Although if `retryFor()` is empty, it it is going to be "retry-n-recovery", like the current behavior:
```
boolean retryNotExcluded = includes.length == 0;
```
which goes to the `defaultValue` of the `SimpleRetryPolicy`.

I guess we can set a new `RetryContext.NO_RECOVERY` attribute in the `SimpleRetryPolicy` for the `noRetryNoRecoveryFor` to avoid polluting `RetryTemplate` API...
The `null` value for those exceptions should be OK in the `Map<Class<? extends Throwable>, Boolean> retryableExceptions` as an indicator for a third state.

PR #265, opened 2022-01-26
Add rethrow flag

ttulka:
this implements #264

### DIFF
--- a/src/main/java/org/springframework/retry/support/RetryTemplate.java
+++ b/src/main/java/org/springframework/retry/support/RetryTemplate.java
@@ -96,6 +96,8 @@ public class RetryTemplate implements RetryOperations {
 
 	private boolean throwLastExceptionOnExhausted;
 
+	private boolean noRecoveryForNotRetryable;
+
 	/**
 	 * Main entry point to configure RetryTemplate using fluent API. See
 	 * {@link RetryTemplateBuilder} for usage examples and details.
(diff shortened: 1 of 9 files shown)


ttulka:
@garyrussell Any updates on this, please?

garyrussell:
Sorry; haven't had a chance to look at it yet; will try this week.

src/main/java/org/springframework/retry/support/RetryTemplate.java
```diff
@@ -529,12 +526,16 @@ private RetryContext doOpenInternal(RetryPolicy retryPolicy) {
...
 	 */
-	protected <T> T handleRetryExhausted(RecoveryCallback<T> recoveryCallback, RetryContext context, RetryState state)
-			throws Throwable {
+	protected <T> T handleRetryExhausted(RecoveryCallback<T> recoveryCallback, RetryContext context, RetryState state,
+			RetryPolicy retryPolicy) throws Throwable {
 		context.setAttribute(RetryContext.EXHAUSTED, true);
 		if (state != null && !context.hasAttribute(GLOBAL_STATE)) {
 			this.retryContextCache.remove(state.getKey());
 		}
+		if (this.throwLastExceptionOnExhausted && retryPolicy instanceof SimpleRetryPolicy
```
garyrussell, line 535: This is a behavior change and could break existing projects that expect the recovery callback to be invoked for all exceptions. I think we need a new boolean `noRecoveryForNotRetryable` or similar.

garyrussell (changes requested):
Thanks; there are a couple of issues; also the Copyright for all affected files needs to be changed to 2022.

src/main/java/org/springframework/retry/support/RetryTemplate.java
ttulka: `throwLastExceptionOnExhausted` is true only if `Retryable.rethrow` is true, which is false per default. So, all the old code will have `throwLastExceptionOnExhausted` set to false and this new `if` branch will be skipped. 

Could you please collaborate more if I miss something? Thanks!

garyrussell: @ttulka Sorry for the delay. The problem is for users of the `RetryTemplate` (those not using `@Retryable`, which is very common).

If an existing application is using the template with `throwLastExceptionOnExhausted` then the recovery callback is still invoked.

With your change, it will no longer be called; hence the new behavior needs to be controlled by a new property on the template rather than overloading the existing property to have 2 meanings.

ttulka: @garyrussell Thanks for clearing things up! I have committed the changes with the new flag.

garyrussell:
@ttulka Thanks for the contribution!

Merged as ec245fd1bac5c392c091757a47863f2769c3e21b and added a few docs 0c1fa828a80f1bafeac55394241929f33c2b9e4a

PR #266, opened 2022-02-14
Add randomExpression to backoff annotation

Deviad:


garyrussell (commented):
Thanks for the contribution. It looks ok to me, but I am curious as to why this needs to be an expression; it seems to me this would be set by the developer and not need to be evaluated during initialization; just interested in the use case.

Deviad:
```
    @Retryable(include = {Exception.class}, backoff = @Backoff(
            delayExpression = "#{new Long(${property.delay})}",
            multiplierExpression = "#{new Long(${property.multiplier})}",
            maxDelayExpression = "#{new Long(${property.maxDelay})}",
            random = "#{new Long(${property.random})}",
    ))
    public FancyDTO somethingFancy(final String foobar) {
    }
```
Instead of creating a feature flag to activate/deactivate it, and use @ConditionalOnProperty, which would cause code duplication (as I would need to copy paste an entire class), I can simply activate/deactivate it with a property set accordingly for each environment (and avoid code duplication). 

This is available already here:
https://github.com/thombergs/code-examples/tree/master/resilience4j/springboot-resilience4j

I managed to reproduce that with SPEL so that props are fetched from properties files.

garyrussell:
Thanks; I understand what it does, and how it does it; my question is **why** someone would want to do it?

Deviad:
Cos we can can monitor the effects of the change as I said (in terms of throughput, back-pressure on the consumers, etc.) and push it forward if it satisfies our needs.

garyrussell (approved):

Issue #276, opened 2022-03-10
Expose BackOffPolicy creation logic through a BackOffPolicyBuilder


tomazfernandes:
This issue is to propose exposing value-based `BackOffPolicy` creation logic, currently in [this part](https://github.com/spring-projects/spring-retry/blob/14c7217bbc70c81ebd5e0ea045410f7e69f788c4/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java#L381) of the `AnnotationAwareRetryOperationsInterceptor` class, through a `BackOffPolicyBuilder` class, so other `Spring` projects and perhaps end users can use it, keeping consistency throughout the projects and avoiding duplicated logic.

**Context**

When `Spring Kafka`'s [Non-blocking retries feature](https://docs.spring.io/spring-kafka/reference/html/#retry-topic) was developed, the `Spring Retry` project wasn't getting many releases. As we wanted to leverage some of `Spring Retry`'s functionality, such as the `@BackOff` annotation and the `BackOffPolicy` interface and implementations, the solution was to duplicate the `BackOffPolicy` creation logic in `Spring Kafka`, as can be seen [here](https://github.com/spring-projects/spring-kafka/blob/f83fc4b11eed2708a488227bccc63d93dc16c3ed/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java#L158).

Now that we're in the process of [adding auto configuration](https://github.com/spring-projects/spring-boot/issues/28450) capabilities to the feature, we need to be able to access this logic in `Spring Boot` as well. After discussing this matter with @wilkinsona and @artembilan, considering the `Spring Retry` project is now getting more support,  it seems like the right approach would be to make this logic accessible through a `BackOffPolicyBuilder` and use it both in `Spring Kafka` and `Spring Boot` projects.

If that does look like the right approach I can provide a PR, if that's ok.

Thanks, and I'm really glad to see this project under more active support again, it's a very useful and well-thought project.

artembilan:
That's what we have discussed, @tomazfernandes .

If you really have something in mind, feel free to raise a PR and I'll review it shortly.

Thanks

tomazfernandes:
Sure @artembilan. As this is a different project I wanted to provide some context to everyone else who might see this issue. I've opened the PR just now.

Thanks

PR #277, opened 2022-03-11
GH-276: Expose BackOffPolicy building logic

tomazfernandes:
`BackOffPolicyBuilder` implementation to expose `BackOffPolicy` building logic to other projects.

A few quick considerations.

My main goals here were consistency with the rest of the project, and reduced risk of breaking behaviors.

That impacted a few decisions, such as:

* I didn't change anything in the actual algorithm to reduce risk of accidentally changing any behavior
* I used `primitives` instead of their classes counterparts (`Long`, `Double`, `Boolean`)
Using classes we'd be able to set a value only if not `null`, delegating to the policy which default value to use. But we'd have to introduce null checks all around, which might interfere the first bullet above. Also, this helps keeping consistency with the rest of the project.
* No validation if the inputs are `> 0`. If these validations were introduced, they would change the behavior for users that may be relying on passing values such as `-1` to the annotation to let the policy set a default value.
* I copied and pasted a lot of the javadocs, besides the main algorithm. I don't know if there's a way to give proper credit in this case.

If you have any suggestions regarding the API, implementation, or anything else, please let me know, and also if there's anything that needs to be changed.

Thanks

### DIFF
--- a/src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
+++ b/src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
@@ -25,6 +25,10 @@ package org.springframework.retry.backoff;
  * <p>
  * Examples: <pre>
  *
+ * // Default {@link FixedBackOffPolicy} with 1000ms delay
+ * BackOffPolicyBuilder
+ * 		.newDefaultPolicy();
+ *
  * // {@link FixedBackOffPolicy}
  * BackOffPolicyBuilder
  * 		.newBuilder()
@@ -52,8 +56,8 @@ package org.springframework.retry.backoff;
  * 		.delay(3000)
  * 		.maxDelay(5000)
  * 		.multiplier(1.5)
- * 		.isRandom(true)
- * 		.withSleeper(mySleeper)
+ * 		.random(true)
+ * 		.sleeper(mySleeper)
  * 		.build();
  * </pre>
  * <p>
@@ -74,7 +78,7 @@ public class BackOffPolicyBuilder {
 
 	private double multiplier;
 
-	private boolean isRandom;
+	private boolean random;
 
 	private Sleeper sleeper;
 
@@ -87,6 +91,14 @@ public class BackOffPolicyBuilder {
 	 */
 	public static BackOffPolicyBuilder newBuilder() {
 		return new BackOffPolicyBuilder();
+	}
+
+	/**
+	 * Creates a new {@link FixedBackOffPolicy} instance with a delay of 1000ms.
+	 * @return the back off policy instance
+	 */
+	public static BackOffPolicy newDefaultPolicy() {
+		return new BackOffPolicyBuilder().build();
 	}
 
 	/**
@@ -125,11 +137,11 @@ public class BackOffPolicyBuilder {
 	 * In the exponential case ({@link #multiplier} &gt; 0) set this to true to have the
 	 * backoff delays randomized, so that the maximum delay is multiplier times the
 	 * previous delay and the distribution is uniform between the two values.
-	 * @param isRandom the flag to signal randomization is required
+	 * @param random the flag to signal randomization is required
 	 * @return this
 	 */
-	public BackOffPolicyBuilder isRandom(boolean isRandom) {
-		this.isRandom = isRandom;
+	public BackOffPolicyBuilder random(boolean random) {
+		this.random = random;
 		return this;
 	}
 
@@ -139,7 +151,7 @@ public class BackOffPolicyBuilder {
 	 * @param sleeper the {@link Sleeper} instance
 	 * @return this
 	 */
-	public BackOffPolicyBuilder withSleeper(Sleeper sleeper) {
+	public BackOffPolicyBuilder sleeper(Sleeper sleeper) {
 		this.sleeper = sleeper;
 		return this;
 	}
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
```diff
@@ -0,0 +1,91 @@
...
+				.isRandom(true).withSleeper(mockSleeper).build();
+		assertTrue(ExponentialRandomBackOffPolicy.class.isAssignableFrom(backOffPolicy.getClass()));
+		ExponentialRandomBackOffPolicy policy = (ExponentialRandomBackOffPolicy) backOffPolicy;
+		assertEquals(10000, policy.getInitialInterval());
+		assertEquals(100000, policy.getMaxInterval());
+		assertEquals(10, policy.getMultiplier(), 0);
+		assertEquals(mockSleeper, new DirectFieldAccessor(policy).getPropertyValue("sleeper"));
+	}
+
+}
```
artembilan, line 91: New line in the end of each file, please.

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * @author Tomaz Fernandes
+ * @since 1.3.3
+ */
+public class BackOffPolicyBuilderTests {
+
+	@Test
+	public void shouldCreateDefaultBackOffPolicy() {
+		Sleeper mockSleeper = mock(Sleeper.class);
+		BackOffPolicy backOffPolicy = BackOffPolicyBuilder.newBuilder().withSleeper(mockSleeper).build();
```
artembilan, line 36: So, looks like it is just possible to have it like this `BackOffPolicyBuilder.newBuilder().build()`?

Why do we need then this `Sleeper` over here?

Can we have then something like `BackOffPolicyBuilder.newDefaultInstance()` for convenience?

But  yeah: we can live so far with what you did already.

src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+	}
+
+	/**
+	 * In the exponential case ({@link #multiplier} &gt; 0) set this to true to have the
+	 * backoff delays randomized, so that the maximum delay is multiplier times the
+	 * previous delay and the distribution is uniform between the two values.
+	 * @param isRandom the flag to signal randomization is required
+	 * @return this
+	 */
+	public BackOffPolicyBuilder isRandom(boolean isRandom) {
```
artembilan, line 131: I think if just `random(boolean)` is enough for a builder method name.

src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
```diff
@@ -0,0 +1,183 @@
...
+		return this;
+	}
+
+	/**
+	 * The {@link Sleeper} instance to be used to back off. Policies default to
+	 * {@link ThreadWaitSleeper}.
+	 * @param sleeper the {@link Sleeper} instance
+	 * @return this
+	 */
+	public BackOffPolicyBuilder withSleeper(Sleeper sleeper) {
```
artembilan, line 142: Why just `sleeper()` doesn't work for a method name?

artembilan (changes requested):
Just a couple minor concerns.

Please, answer or fix and we will merge shortly.

Thanks

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
tomazfernandes: > So, looks like it is just possible to have it like this `BackOffPolicyBuilder.newBuilder().build()`?

Yes

> Why do we need then this `Sleeper` over here?

I added the sleeper to all tests because the sleeper addition logic is repeated for all policies. So, doesn't hurt to make sure it's working in all cases, including the default one. I can remove it, specially if we're creating a dedicated method for the default policy.

> Can we have then something like `BackOffPolicyBuilder.newDefaultInstance()` for convenience?

Sure. I think maybe `newDefaultInstance` might give the impression that it's a new default `Builder` instance. Maybe `newDefaultPolicy` instead?

> But yeah: we can live so far with what you did already.

That's nice, thanks.

artembilan: Sure! The `newDefaultPolicy()` sounds better.

Thanks

tomazfernandes:
> Just a couple minor concerns.
> 
> Please, answer or fix and we will merge shortly.
> 
> Thanks

Ok, I should have it ready in a couple of hours - have an appointment now. I'll also change the method names to the ones you suggested.

Thanks a lot for the quick review.

tomazfernandes:
Actually my appointment is not until the next hour 😄

I've updated the PR. If there is anything you think we could improve on let me know, I can either work further on this PR or open a new one after this one is merged.

Thanks

artembilan:
Thanks, @tomazfernandes ; looking forward for more!

tomazfernandes:
> Thanks, @tomazfernandes ; looking forward for more!

You're welcome @artembilan, thanks for the quick response and support!

Issue #231, opened 2021-02-01
Try again  base on the result not just exception


githubcheng2978:
spring-retry  support to try again  according to the value of the exception, but not support the execution result of the doWithRetry method

dsyer:
That's a pretty big change to the design. No-one ever asked for it, so I think we'll wait and see if anyone else is interested.

PR #289, opened 2022-04-19
GH-231: Support Retry Based on Result

garyrussell:
Resolves https://github.com/spring-projects/spring-retry/issues/231

Enhance `RetryListener` to allow it to examine the result to determine
whether the call should be accepted, or retried according to the retry
policy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,7 +362,7 @@ The type of the exception thrown is then used to determine whether the call shou
 If there has been an error, it is the last one thrown by the `RetryCallback`.
 
 Note that when there is more than one listener, they are in a list, so there is an order.
-In this case, `open` is called in the same order, while `onSuccess`, `onError`, and `close` are called in reverse order
+In this case, `open` is called in the same order, while `onSuccess`, `onError`, and `close` are called in reverse order.
 
 `RetryListenerSupport` is provided, with no-op implementations; you can extend this class if you don't need to implement all of the `RetryListener` methods.
 
--- a/README.md
+++ b/README.md
@@ -345,26 +345,34 @@ public interface RetryListener {
 
     void open(RetryContext context, RetryCallback<T> callback);
 
+    void onSuccess(RetryContext context, T result);
+
     void onError(RetryContext context, RetryCallback<T> callback, Throwable e);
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -345,27 +345,35 @@ public interface RetryListener {
...
+This allows the listener to examine the result and throw an exception if the result doesn't match some expected criteria.
+The type of the exception thrown is then used to determine whether the call should be retried or not, based on the retry policy.
+If there has been an error, it is the last one thrown by the `RetryCallback`.
 
 Note that when there is more than one listener, they are in a list, so there is an order.
-In this case, `open` is called in the same order, while `onError` and `close` are called
-in reverse order.
+In this case, `open` is called in the same order, while `onSuccess`, `onError`, and `close` are called in reverse order
+
+`RetryListenerSupport` is provided, with no-op implementations; you can extend this class if you don't need to implement all of the `RetryListener` methods.
```
artembilan, line 367: I don't think we need this one from now on since we just simply can go ahead with `default` methods in the interface.
Perhaps separate issue though...

src/main/java/org/springframework/retry/support/RetryTemplate.java
```diff
@@ -590,6 +592,13 @@ private <T, E extends Throwable> void doCloseInterceptors(RetryCallback<T, E> ca
 		}
 	}
 
+	private <T, E extends Throwable> void doOnSuccessInterceptors(RetryCallback<T, E> callback, RetryContext context,
+			T result) {
+		for (int i = this.listeners.length; i-- > 0;) {
```
artembilan, line 597: Is there a clear view why we visit them in the reverse order?
I know: this comes as copy/paste from the `doOnErrorInterceptors()`, but still - "why?"

Thanks

artembilan (changes requested):

src/main/java/org/springframework/retry/support/RetryTemplate.java
garyrussell: I don't know the reason; but it's clearly stated in the README.

README.md
garyrussell: True, but we'd have to go through a deprecation cycle; definitely a separate issue.

PR #290, opened 2022-04-20
Fix to use multiple recovery method in same class

praveensam123:
Usage: @Retryable(recover="methodName")


artembilan (changes requested):
Can you confirm the problem, please, with some unit test?
For example in the `RecoverAnnotationRecoveryHandlerTests`

artembilan:
This looks like a duplication of https://github.com/spring-projects/spring-retry/pull/371.
Will be released in next version this week.

Thank you for your effort!

Issue #184, opened 2019-10-25
Annotation Improvements


garyrussell:
Evaluate `*Expression` properties at runtime instead of during annotation parsing.

See https://stackoverflow.com/questions/58559733/can-springs-retryable-or-retrytemplate-use-retry-after-header-for-dynamic-backo/58560029#58560029

We could either enhance the current policies and backoff policies to  accept expressions directly or, after moving to 2.0 (and Java 8) `Supplier<?>`s and leave the expression evaluation in the `AnnotationAwareRetryOperationsInterceptor`.

PR #298, opened 2022-05-03
GH-184: Expression Evaluation at Runtime

garyrussell:
Resolves https://github.com/spring-projects/spring-retry/issues/184


src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
```diff
@@ -218,7 +240,7 @@ private boolean retryForException(Throwable ex) {
 
 	@Override
 	public String toString() {
-		return ClassUtils.getShortName(getClass()) + "[maxAttempts=" + this.maxAttempts + "]";
+		return ClassUtils.getShortName(getClass()) + "[maxAttempts=" + this.maxAttemptsSupplier + "]";
```
artembilan, line 243:
```suggestion
		return ClassUtils.getShortName(getClass()) + "[maxAttempts=" + getMaxAttempts() + "]";
```
?

src/main/java/org/springframework/retry/backoff/UniformRandomBackOffPolicy.java
```diff
@@ -107,9 +132,10 @@ public long getMaxBackOffPeriod() {
 	 */
 	protected void doBackOff() throws BackOffInterruptedException {
 		try {
-			long delta = maxBackOffPeriod == minBackOffPeriod ? 0
-					: random.nextInt((int) (maxBackOffPeriod - minBackOffPeriod));
-			sleeper.sleep(minBackOffPeriod + delta);
+			Long min = this.minBackOffPeriod.get();
+			long delta = this.maxBackOffPeriod == this.minBackOffPeriod ? 0
```
artembilan, line 136: I think this is something wrong: you compare `Supplier` instances - not their values.

artembilan (changes requested):
I'm not fully buying this evaluation mode and would live with just a runtime one like we do in many other places. Plus it might turn in some use-case that end-user would like to have a mix of modes for different expressions. Also it looks a bit clumsy for nested annotations...

src/main/java/org/springframework/retry/backoff/UniformRandomBackOffPolicy.java
garyrussell: Oops.

garyrussell:
I don't think we should add the overhead of runtime evaluation for all users; when most won't need it; it also causes problems because the `SimpleRetryPolicy` must be serializable (used in the circuit breaker retry context. See the note in the README and javadocs.

garyrussell:
One alternative I considered was template expressions `#{...}` are evaluated at initialization and non-template expressions are evaluated at runtime.

This would be a behavior change, but it's a major release, so I could agree to it.

artembilan:
Yeah...

If we would like to be consistent with an XML configuration and its SpEL support, then indeed that `@Retryable(maxAttempts)` has to be a string and support a "templated" expression evaluation during configuration phase.
Meanwhile a `maxAttemptsExpression` would really lead to the runtime expression evaluation. 
Just similar to what we have so far in Spring Integration.

Looking to the same  `@Retryable` right now we have:
```
	int maxAttempts() default 3;

	String maxAttemptsExpression() default "";

	String exceptionExpression() default "";
```

Where it is not clear that `maxAttemptsExpression` is going to be evaluated at configuration phase or at runtime, but `exceptionExpression` states it clearly that it is for runtime.
That's another argument from me against a common  evaluation mode option for the whole annotation.

So, if we cannot change simple value attributes (like that `maxAttempts` to String), then your suggestion for templating on companion SpEL attributes doesn't sound bad to distinguish runtime and configuration phase evaluations.

garyrussell:
I think changing the types would be too much of a breaking change - even for a major release (I suspect many users don't even use expressions); I'll go with templating idea.

src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
```diff
@@ -149,11 +153,29 @@ public void setMaxAttempts(int maxAttempts) {
...
 
+	/**
+	 * Set a supplier for the number of attempts before retries are exhausted. Includes
+	 * the initial attempt before the retries begin so, generally, will be {@code >= 1}.
+	 * For example setting this property to 3 means 3 attempts total (initial + 2
+	 * retries). IMPORTANT: This policy cannot be serialized when a max attempts supplier
+	 * is provided. Serialization might be used by a distributed cache when using this
+	 * policy in a {@code CircuitBreaker} context.
+	 * @param maxAttemptsSupplier the maximum number of attempts including the initial
+	 * attempt.
```
artembilan, line 164: `@since 2.0` ?

src/main/java/org/springframework/retry/policy/SimpleRetryPolicy.java
artembilan: Looks like has not been addressed.

README.md
```diff
@@ -588,7 +588,33 @@ Expressions can contain property placeholders, such as `#{${max.delay}}` or
...
 - `maxAttemptsExpression` and the `@BackOff` expression attributes are evaluated once,
 during initialization. There is no root object for the evaluation but they can reference
-other beans in the context.
+other beans in the context
+
+Starting with version 2.0, expressions in `@Retryable`, `@CircuitBreaker`, and `BackOff` can be evaluated once, during application initialization, or at runtime.
+With earlier versions, evaluation was always performed during initialization (except for `Retryable.exceptionExpression` which is always evaluated at runtime).
+When evaluating at runtime, a root object containing the method arguments is passed to the evaluation context.
+
+**Note:** The arguments are not available until the method has been called at least once; they will be null initially, which means, for example, you can't set the initial `maxAttempts` using an argument value, you can, however, change the `maxAttempts` after the first failure and before any retries are performed.
```
artembilan, line 597: Is it only a limitation for the `maxAttempts` option?

I see `open()` is done far before a `doWithRetry()`, so perhaps those `args` does not make sense for circuit breaker...

Looks like BackOff expressions don't suffer from this limitation: it is indeed called in the `catch` from the `doWithRetry()`.

Do I miss anything?

Thanks

artembilan (changes requested):

README.md
garyrussell: It turns out that we can't use args at all for any stateful retry (including circuit breakers) because you would get the args from the previous call (which may be different).

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -476,14 +477,16 @@ private boolean isTemplate(String expression) {
...
+		if (stateless) {
+			RetryContext context = RetrySynchronizationManager.getContext();
+			if (context != null) {
+				args = (Args) context.getAttribute("ARGS");
+			}
+			if (args == null) {
+				args = Args.NO_ARGS;
+			}
 		}
 		return expression.getValue(this.evaluationContext, args, type);
```
artembilan, line 491: Is it expected to propagate an `args`  var as `null` for statefull unlike for stateless with an `Args.NO_ARGS`?

artembilan (changes requested):

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
garyrussell: Yes; I want them to get a hard error `EL1007E: Property or field 'args' cannot be found on null`.

I failed to save the doc change for this; pushing now...

Issue #299, opened 2022-05-18
IllegalArgumentException when migrating from 1.3.1 to 1.3.2 and using class annotations


deadpassive:
Hi,

I am using a setup similar to the following, where I have a `@Retryable` defined on the class and also a `@Recover` method.

```
@Retryable(value = {MyFirstException.class, MySecondException.class})
@Service
public class MyService {

  public void myMethod(String myArg) {
    System.out.println("Doing a thing that will fail with args " + myArg);
    throw new MyFirstException();
  }

  @Recover
  public void recoverMyMethod(Exception e, String myArg) throws Exception {
    System.out.println("In recovery, with args " + myArg);
    throw e;
  }
}
```

When running with spring-retry v1.3.1 this worked fine and `myMethod` is run 3 times before `recoverMyMethod` is called a single time before the exception is finally thrown.

When running with v1.3.2 or v1.3.3 of spring-retry the behaviour changes and it also retries the recover method too. So `myMethod` is called 3 times and then `recoverMyMethod` is called 3 times before finally an `IllegalArgumentException` is thrown. 

I think this exception is thrown because it is trying to call `recoverMyMethod` to recover from itself and gets confused when creating the args to call it and ends up trying to call it with `(Exception, Exception)` instead of `(Exception, String)`.

Is this a bug in spring-retry, or are we using it incorrectly?

Here is an example project which illustrates the issue:
[spring-retry-test.zip](https://github.com/spring-projects/spring-retry/files/8714549/spring-retry-test.zip)

garyrussell:
I am not sure (yet) what caused it, but the behavior has definitely changed; as a work-around, moving the `@Retryable` to the method (instead of declaring it at the class level) works.

garyrussell:
It was broken by the fix for this https://github.com/spring-projects/spring-retry/issues/244

garyrussell:
Another work around is to remove the `public` modifier from the recover method; that prevents the interceptor from proxying that method.

PR #300, opened 2022-05-18
GH-299: Fix @Recover with Class Level @Retryable

garyrussell:
Resolves https://github.com/spring-projects/spring-retry/issues/299

Caused by the fix for #244, recover methods should not be considered
as retryable, with class level annotation.

**cherry-pick to 1.3.x**


artembilan:
Back-ported as https://github.com/spring-projects/spring-retry/commit/7063d363862636ece7a8bf6a2b7b781589320dff after fixing conflicts. Mostly JUnit 5 -> JUnit 4 migration back.

PR #326, opened 2022-11-23
Add default implementations to `RetryListener`

hpoettker:
With Spring Retry 2.0.0, the baseline was lifted from Java 6 to Java 17, which allows to now use default implementations in interfaces.

The only purpose of `RetryListenerSupport` is to provide default implementations for `RetryListener`. As such support implementations are no longer required, this PR deprecates `RetryListenerSupport` and adds its default implementations to `RetryListener`.

### DIFF
--- a/src/main/java/org/springframework/retry/listener/RetryListenerSupport.java
+++ b/src/main/java/org/springframework/retry/listener/RetryListenerSupport.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2007 the original author or authors.
+ * Copyright 2006-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,8 +24,10 @@ import org.springframework.retry.RetryListener;
  * Empty method implementation of {@link RetryListener}.
  *
  * @author Dave Syer
- *
+ * @author Henning Pöttker
+ * @deprecated in favor of the default implementations in {@link RetryListener}
  */
+@Deprecated(since = "2.0.1", forRemoval = true)
 public class RetryListenerSupport implements RetryListener {
 
 	public <T, E extends Throwable> void close(RetryContext context, RetryCallback<T, E> callback,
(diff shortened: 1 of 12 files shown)


src/test/java/org/springframework/retry/listener/RetryListenerSupportTests.java
```diff
@@ -21,6 +21,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatNoException;
 
+@SuppressWarnings("deprecation")
```
artembilan, line 24: I think it is better to rework tests in this class to deal with new code base.
I believe that this `RetryListenerSupport` will be removed in the next `2.1` version.
But that doesn't mean we will remove this test class.
So, let's see if just `new RetryListener()` is enough for unit tests in this class!

src/main/java/org/springframework/retry/listener/RetryListenerSupport.java
```diff
@@ -24,8 +24,9 @@
  * Empty method implementation of {@link RetryListener}.
  *
  * @author Dave Syer
- *
+ * @deprecated in favor of the default implementations in {@link RetryListener}
  */
+@Deprecated
```
artembilan, line 29: `since="2.0.1", forRemoval=true` ?

artembilan (changes requested):
Please, add your name to the `@author` list of all the affected classes.
Also update their copyright to the current `2022`.

There is also a mentioning of `RetryListenerSupport` in the README.
Consider, please, revise that docs respectively.
(Not sure if we should mention its deprecation: just remove will be enough)

src/main/java/org/springframework/retry/listener/RetryListenerSupport.java
hpoettker: done

src/test/java/org/springframework/retry/listener/RetryListenerSupportTests.java
hpoettker: There's already a class `RetryListenerTests`. I added the corresponding tests over there, and left this class as is and ready to be removed.

hpoettker:
@artembilan Thanks for the feedback!

The year in the headers is already 2022 in all affected classes. But I've added the `@author` tags. The line from the README is also removed.

artembilan:
@hpoettker ,

PR #330, opened 2023-02-08
add notice and license to jar

tobi5775:
In many other Spring and Spring Boot projects, the license and notice files are included in the jar file under META-INF. In this Spring project, the two files should also be added to the jar.

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2006-2022 the original author or authors.
+Copyright (c) 2006-2023 the original author or authors.
 
 This product is licensed to you under the Apache License, Version 2.0
 (the "License"). You may not use this product except in compliance with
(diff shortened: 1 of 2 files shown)


NOTICE.txt
```diff
@@ -0,0 +1,5 @@
+Copyright (c) 2006-2022 the original author or authors.
```
artembilan, line 1: Probably `2023` already.
May I know, please, from what project did you take this notice?
I see that in my other one it is slightly different.
Thanks

artembilan (changes requested):

NOTICE.txt
tobi5775: I tooked the NOTICE File from the [spring-boot](https://github.com/spring-projects/spring-boot/blob/main/buildSrc/src/main/resources/NOTICE.txt) project.

artembilan:
Thank you, @tobi5775 , for contribution; looking forward for more!

Issue #22, opened 2015-01-06
support custom RetryConfiguration.getOrder() via @EnableRetry like @EnableAsync


quaff:
``` java
@EnableRetry(order = -1)
@Configuration
public class RetryConfiguration {

}

```

configure @Retryable before @Transactional, so we can handle transaction with stateless @Retryable,  is stateful necessary anymore?

headstar:
+1

headstar:
Is it correct to say that the transactional advisor and retryable advisor are unordered in the snippet below? (using @EnableRetry).
```
@Transactional
@Retryable
@Service
public class FooService {
}
```

quaff:
Yes, you can do this
```xml
<tx:annotation-driven order="0"/>
```
```java
@Configuration
public class RetryConfiguration extends
		org.springframework.retry.annotation.RetryConfiguration {

	@Override
	public int getOrder() {
		return -1;
	}

}
```

headstar:
Ok, thanks.

aartigao:
Hey, I'm looking for this feature. Are there any plans on implement it?

quaff:
ping

DieterVDW:
Pong!

Is this project still being maintained?

ejust:
+ 1

yoorick:
+1 for this feature.

We have multiple advisors in our project, and a `MethodInvocation` usually consists of a chain of two or more `MethodInterceptor`s, and a particular order of them is important.

Now we have to use an approach similar to what @quaff showed above.

wtritch:
In case others are still having this issue, we ran into a problem using a `@Backoff` annotation in a `@Retryable @Transactional`-annotated method.  The `@Backoff` annotation includes an `@Import(RetryConfiguration.class)` which leads to double-registering retry AOP Advice (one for the class extending RetryConfiguration, and one for the imported RetryConfiguration.)

After some mucking around with Spring's configuration loading we decided on manually setting the RetryConfiguration order in a `@PostConstruct` which needs to be executed PRIOR to instantiating any advised beans.  (Once a bean is advised the advice order is locked.)
```
@Configuration
@Order(Ordered.HIGHEST_PRECEDENCE)
@EnableRetry
public class OrderRetryConfiguration {

    @Autowired
    public RetryConfiguration retryConfiguration;

    @PostConstruct
    public void postConstruct() {
        this.retryConfiguration.setOrder(Ordered.LOWEST_PRECEDENCE - 1);
[...]

PR #333, opened 2023-02-21
Introduce order for @EnableRetry

quaff:
Fix GH-22

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/EnableRetry.java
+++ b/src/main/java/org/springframework/retry/annotation/EnableRetry.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import java.lang.annotation.Target;
 
 import org.springframework.context.annotation.EnableAspectJAutoProxy;
 import org.springframework.context.annotation.Import;
+import org.springframework.core.Ordered;
 import org.springframework.core.annotation.AliasFor;
 
 /**
@@ -33,7 +34,8 @@ import org.springframework.core.annotation.AliasFor;
  * the annotations.
  *
  * @author Dave Syer
- * @since 2.0
+ * @author Yanming Zhou
+ * @since 1.1
  *
  */
 @Target(ElementType.TYPE)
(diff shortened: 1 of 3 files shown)


artembilan (approved):

artembilan:

src/main/java/org/springframework/retry/annotation/EnableRetry.java
```diff
@@ -51,4 +53,13 @@
 	@AliasFor(annotation = EnableAspectJAutoProxy.class)
 	boolean proxyTargetClass() default false;
 
+	/**
+	 * Indicate the order in which the {@link RetryConfiguration} should be applied.
+	 * <p>
+	 * The default is {@link Ordered#LOWEST_PRECEDENCE} in order to run after all other
+	 * post-processors, so that it can add an advisor to existing proxies rather than
+	 * double-proxy.
```
xak2000, line 61: Looks like there is an error in this description. The reasoning is misleading. The `RetryConfiguration` advice is not applied by any post-processors AFAIK. It is just registered as a bean to be applied later by bean-factory when it creates a proxy for a bean. And it's always applied just once, so there is no danger of creating double-proxy.

@artembilan Any chance this description could be corrected? To be honest I don't know the real reason of this property to be `Ordered.LOWEST_PRECEDENCE` except the backward compatibility. This default is actually danger as it interferes with `@Transactional` annotation that is also `Ordered.LOWEST_PRECEDENCE`, so the final order is undefined. But 99% of the time you want the order to be `@Retry -> @Transactional` and not vice versa because many exceptions will be thrown only at the commit phase and could be retried only by starting a new transaction.

artembilan: Sure! We can fix that description and also make a default order as a `Ordered.LOWEST_PRECEDENCE - 1` to make it working with `@Transactional` as you have just explained which is really makes sense as a default behavior.
I think that description was taken from the `@EnableAsync` which looks like does its logic slightly different than an `AspectJAwareAdvisorAutoProxyCreator` initiated by the `@EnableAspectJAutoProxy` from this `@EnableRetry`.

Feel free to contribute the fix!
Thank you!

quaff:
@artembilan Could you backport this and https://github.com/spring-projects/spring-retry/pull/335 to 1.3.x? many projects doesn't upgrade JDK to 17.

artembilan:
Here it is: https://github.com/spring-projects/spring-retry/issues/372.

I'll see what and how we can back-port it.

Thanks

PR #335, opened 2023-03-08
Change default order for @EnableRetry

xak2000:
The new default order is `Ordered.LOWEST_PRECEDENCE - 1` to make sure the advice is applied before other advices with
`Ordered.LOWEST_PRECEDENCE` order (e.g. an advice responsible for `@Transactional` behavior).

See also https://github.com/spring-projects/spring-retry/pull/333#pullrequestreview-1330977780


artembilan (changes requested):
Please, run `mvnw spring-javaformat:apply` and push a new commit.

PR #343, opened 2023-05-23
Add constructor make timeout customisable

aahlenst:
Adds a separate constructor to `TimeoutRetryPolicy` to create a new instance with a configurable timeout. This follows the pattern established by other policies, for example, `SimpleRetryPolicy`, that allow to customise their behaviour during instance creation.


artembilan (changes requested):

aahlenst:
Problems should be resolved.

artembilan:

PR #344, opened 2023-05-25
Add support of Duration in RetryTemplateBuilder

aahlenst:
Introduce java.time.Duration as an alternative to express timeouts and intervals when building RestTemplate.

### DIFF
--- a/src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
+++ b/src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
@@ -157,7 +157,7 @@ public class RetryTemplateBuilder {
 	 * @since 2.0.2
 	 */
 	public RetryTemplateBuilder withTimeout(Duration timeout) {
-		Assert.notNull(timeout, "timeout is null");
+		Assert.notNull(timeout, "timeout must not be null");
 		return withTimeout(timeout.toMillis());
 	}
 
@@ -225,8 +225,8 @@ public class RetryTemplateBuilder {
 	 * @since 2.0.2
 	 */
 	public RetryTemplateBuilder exponentialBackoff(Duration initialInterval, double multiplier, Duration maxInterval) {
-		Assert.notNull(initialInterval, "initialInterval is null");
-		Assert.notNull(maxInterval, "maxInterval is null");
+		Assert.notNull(initialInterval, "initialInterval must not be null");
+		Assert.notNull(maxInterval, "maxInterval must not be null");
 		return exponentialBackoff(initialInterval.toMillis(), multiplier, maxInterval.toMillis(), false);
 	}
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
```diff
@@ -132,6 +133,20 @@ public RetryTemplateBuilder withinMillis(long timeout) {
...
+	/**
+	 * Retry until {@code timeout} has passed since the initial attempt.
+	 * @param timeout duration for how long retries should be attempted
+	 * @return this
+	 * @see TimeoutRetryPolicy
+	 * @throws IllegalArgumentException if timeout is {@code null} or 0, or if another
+	 * retry policy has already been configured.
+	 * @since 2.0.2
+	 */
+	public RetryTemplateBuilder withTimeout(Duration timeout) {
```
aahlenst, line 145: I found `withinMillis()` not to be suitable as a method name if the argument is a `java.time.Duration`. Therefore, I picked `withTimeout()`.

I suggest to deprecate `withinMillis` and to add a `withTimeout(long)` instead. Can do this in a follow-up PR.

artembilan: Sure! What you are saying totally makes sense and you are free to do that deprecation directly in this PR.
Also, please, add your name to the `@author` of the affected classes.

artembilan (changes requested):

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
aahlenst: Is it acceptable to deprecate a method in a patch release? It might break the build for people that treat warnings as errors. Might be less disruptive to do that in a 2.1.0. My question would then be how to proceed. Change `@since` to 2.1.0?

artembilan: Yes, it is acceptable.
If you upgrade and your build fails for those deprecations, then you'd just need to fix them.
Luckily we will have an easy pass to migrate.
Then in the next version we might just simply to remove this deprecated method.

I my feeling it is not OK to treat a deprecation warning as an error.

aahlenst:
I applied the requested changes and deprecated `withinMillis()`.

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
```diff
@@ -132,6 +136,36 @@ public RetryTemplateBuilder withinMillis(long timeout) {
...
+		Assert.notNull(timeout, "timeout is null");
+		return this.withinMillis(timeout.toMillis());
```
artembilan, line 166: No `this.` on method calls, please.
And why do you call a deprecated method from here then?

artembilan (changes requested):

artembilan (changes requested):

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
```diff
@@ -122,16 +124,43 @@ public RetryTemplateBuilder maxAttempts(int maxAttempts) {
...
+	 * Retry until {@code timeout} has passed since the initial attempt.
+	 * @param timeout duration for how long retries should be attempted
+	 * @return this
+	 * @see TimeoutRetryPolicy
+	 * @throws IllegalArgumentException if timeout is {@code null} or 0, or if another
+	 * retry policy has already been configured.
+	 * @since 2.0.2
+	 */
+	public RetryTemplateBuilder withTimeout(Duration timeout) {
+		Assert.notNull(timeout, "timeout is null");
```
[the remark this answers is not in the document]
aahlenst: Better?

artembilan:
Thank you for contribution, @aahlenst ; looking forward for more!

PR #345, opened 2023-05-25
Migrate building PRs to GitHub Action

artembilan:
* Add GH action for PR builds
* Upgrade dependencies and plugins
* Fix deprecations


snicoll:
@artembilan is there a reason with this single commit is doing so many different things? IMO, this PR should only migrate building PRs to GH and remove the feature on our CI.

snicoll:
I've merged this PR, reducing the scope to the PR handling.

snicoll:
I've also backported this to 1.3.x, see https://github.com/spring-projects/spring-retry/commit/5146d0ec914f2969a72135a19b536b2410a97009

artembilan:
> is there a reason with this single commit is doing so many different things?

Well, no one does that housekeeping in this project and that was just a reasonable versions upgrade without any breaking changes.

I find it as too much routine to have individual issues and commits just for version numbers.
Is there a reason behind such a fine-grained management ?

I also see that you have missed an upgrade to SF `6.0.9` which was really the first thing I did in that pom.

Plus you see I have fixed deprecation warnings.
I assume you'd be OK if I push such a fix directly into `main`, but I thought it would be great to have it in the PR for possible discussion around.

And it feels a bit awkward to have a dedicated issue and commit just to fix Dave's email in `dev` section of the pom 😄 

Thank for merging this at least some way and for follow up clean ups!

snicoll:
>  Is there a reason behind such a fine-grained management ?

Regular git and GitHub best practices. If an issue is about updating the CI pipeline, it shouldn't touch unrelated code or upgrade dependencies. This looks pretty sane to me.

> I also see that you have missed an upgrade to SF 6.0.9 which was really the first thing I did in that pom.

I did not. It wasn't relevant to the PR. I did upgrade some plugins as they were leading to warnings in Maven 3.9.2, which was triggered by some of the change but unrelated to the PR.

> Plus you see I have fixed deprecation warnings.

Yep. Those can be pushed with a polish commit, sure.

> And it feels a bit awkward to have a dedicated issue and commit just to fix Dave's email in dev section of the pom 😄

Unless I am missing something, I didn't infer that it should be the case.

artembilan:
> it shouldn't touch unrelated code or upgrade dependencies

OK. I will try to follow this rule next time.

> I did not. It wasn't relevant to the PR.

Sure! I got it now. Then we will have something like `Upgrade dependencies` commit close to the release. It is just too much burden with issues just for a single version number...

> Those can be pushed with a polish commit, sure.

I will create a respective issue and push commit for deprecation warnings.

>  I didn't infer that it should be the case.

Unless I didn't understand this English phrase I believe that info in the project has to be updates as a part of housekeeping, but having a dedicated issue for just email is too much 🤷

snicoll:
I don't think you need an issue for polish commit so the deprecation warnings can go with a polish commit. And for Dave's email that's polishing as well so just a commit is perfectly fine IMO.

PR #363, opened 2023-06-16
Fix link to RetryTemplate examples

raimon49:
Hey, thanks for maintaining a great product.

I was reading the README document and found the internal link is not working.

`<a name="">` style link is no longer supported by many web browsers as of 2023.

This PR is use internal links naturally generated by GitHub instead of setting internal links with your own HTML tags.


snicoll:
Good catch, thanks @raimon49

PR #375, opened 2023-07-06
Option to Avoid ThreadLocal Usage (Draft)

garyrussell:
* Add an option to store contexts in a map instead of a `ThreadLocal`
* Overloaded `execute` methods to accept a unique context key
* Propagate context to runtime expression evaluations (for method argument access)
* Subclass existing tests to verify functionality

Draft only, for review; need to subclass many more tests.

### DIFF
--- a/src/main/java/org/springframework/retry/RetryContext.java
+++ b/src/main/java/org/springframework/retry/RetryContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2022 the original author or authors.
+ * Copyright 2006-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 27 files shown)


src/main/java/org/springframework/retry/RetryContext.java
```diff
@@ -17,16 +17,25 @@
 package org.springframework.retry;
 
 import org.springframework.core.AttributeAccessor;
+import org.springframework.retry.support.RetrySynchronizationManager;
```
artembilan, line 20: Not used in the code: just Javadoc, so FQCN instead to avoid package tangle.

src/main/java/org/springframework/retry/RetryOperations.java
```diff
@@ -17,13 +17,15 @@
 package org.springframework.retry;
 
 import org.springframework.retry.support.DefaultRetryState;
+import org.springframework.retry.support.RetrySynchronizationManager;
```
artembilan, line 20: DITTO about explicit import for just Javadocs.

src/main/java/org/springframework/retry/RetryOperations.java
```diff
@@ -76,6 +116,34 @@ <T, E extends Throwable> T execute(RetryCallback<T, E> retryCallback, RecoveryCa
...
+	 * See implementations for configuration details.
+	 * @param retryCallback the {@link RetryCallback}
+	 * @param retryState the {@link RetryState}
+	 * @param <T> the type of the return value
+	 * @param <E> the type of the exception to return
+	 * @return the value returned by the {@link RetryCallback} upon successful invocation,
+	 * and that returned by the {@link RecoveryCallback} otherwise.
+	 * @param contextKey a key to use to map the context when
+	 * {@link RetrySynchronizationManager#isUseThreadLocal()} is false. Must be unique for
+	 * each execution.
```
artembilan, line 136: Cannot we have it randomly generated then instead?
Or something like `contextKeySupplier` injected into a `RetryTemplate`...

artembilan (changes requested):
Just a couple concerns.
Let's discuss with other team members!

src/main/java/org/springframework/retry/RetryOperations.java
garyrussell: Good idea.

garyrussell:
I think I found a much simpler solution. Expect a new PR after VMware Explore.

Issue #211, opened 2020-07-17
Provide The Ability to Exclude Global RetryListeners


krooj:
Hey,

I recently encountered this case whereby methods which are annotated `@Retryable` with no list of `listeners` being passed automatically has them added to the executing `RetryTemplate` at runtime. Upon reading up on the source, this is expected, since `RetryListener` instances which are injected into the application context become, so called, "global" listeners. Unfortunately, this has unexpected consequences in some cases, and it would be awesome to also be able to specify a list of `excludeListeners` in the `@Retryable` annotation. 

I know it's feasible to also pass a label and introspect it at the listener level to make the execution decision, but that, afaik, isn't the intended purpose of that parameter.

garyrussell:
A simple work around would be to add

```java
@Bean
RetryListener noOpListener() {
    return new RetryListenerSupport();
}
```

and `listeners = "noOpListener"`.

krooj:
> A simple work around would be to add
> 
> ```java
> @Bean
> RetryListener noOpListener() {
>     return new RetryListenerSupport();
> }
> ```
> 
> and `listeners = "noOpListener"`.

It's a little funky and will require documentation to the effect... I'll give it a go, but my initial reaction was to migrate to declarative code.

ssjf409:
May I handle this issue? please assign to me~ I'll create PR soon!

artembilan:
Well, this is what is said on that option:
```
	/**
	 * Bean names of retry listeners to use instead of default ones defined in Spring
	 * context
	 * @return retry listeners bean names
	 */
	String[] listeners() default {};
```

I suggest to make ignoring global listeners is to modify the logic to react to something like this: `listeners = ""`.
Of course, with the meaning "no listeners at all".

Right now it fails like:
```
org.springframework.beans.factory.NoSuchBeanDefinitionException: No bean named '' available
```
So, that `AnnotationAwareRetryOperationsInterceptor.getListenersBeans(String[] listenersBeanNames)` has to be modified to react to the empty string.
This is going to be a new behavior without breaking change an existing one where we don't have a `listeners`, therefore rely on global configuration.

@ssjf409 ,

Feel free to provide a contribution, but we assign issues only to team members.

Thanks

ssjf409:
Hi. akenra. I saw your PR.
It was nice PR.

Before I created PR, I thought two options 

I came up with two different implementations of @artembilan's suggestion.
[...]

PR #392, opened 2023-10-21
GH-211: Provide the ability to exclude global retryListeners

akenra:
Hey folks,
I was evaluating this library in detail for the use in our enterprise project and came across an issue that seemed rather trivial to carry out an implementation for as per @artembilan's suggestion here: https://github.com/spring-projects/spring-retry/issues/211#issuecomment-1761945504. 

To the best of my knowledge this implementation is fully reverse-compatible.

Please refer to GH-211 for more context if necessary.

upd: made a few corrections that felt right after a good night's sleep 😄

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/Retryable.java
+++ b/src/main/java/org/springframework/retry/annotation/Retryable.java
@@ -167,8 +167,8 @@ public @interface Retryable {
 	/**
 	 * Bean names of retry listeners to use instead of default ones defined in Spring
 	 * context. If this attribute is set to an empty string {@code ""}, it will
-	 * effectively exclude all retry listeners, including with the default listener beans, from being
-	 * used.
+	 * effectively exclude all retry listeners, including with the default listener beans,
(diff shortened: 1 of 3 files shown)


artembilan (approved):
Sorry, I've this somehow.
Looks good to me.

Let's hear @garyrussell 's opinion and have it merged for the next scheduled version!

src/main/java/org/springframework/retry/annotation/Retryable.java
```diff
@@ -165,7 +166,9 @@
 
 	/**
 	 * Bean names of retry listeners to use instead of default ones defined in Spring
-	 * context
+	 * context. If this attribute is set to an empty string {@code ""}, it will
+	 * effectively exclude all retry listeners, along with the default ones, from being
```
garyrussell, line 170:
```suggestion
	 * effectively exclude all retry listeners, including with the default listener beans, from being
```

garyrussell (approved):

Issue #395, opened 2023-10-31
Expose the number of max attempts to the retry context for all policies


e-ivaldi:
It is sometimes useful to be aware, within a retry context, whether the current attempt is the last attempt when using a 'maximum number of attempts before failure' type of retry policy.

Following up on this [discussion](https://github.com/spring-projects/spring-retry/discussions/362) I am opening a small PR to add that type of support to spring-retry.

In order to achieve the desired result, for all policies, a new attribute is going to be added to the context named "context.max-attempts" that will contain the number of max attempts set in the policy, if any, or -1 otherwise.

The PR won't provide extra support to read and parse the attribute value via ad-hoc methods added to the retry contexts themselves.

PR link: https://github.com/spring-projects/spring-retry/pull/396

deathcoder:
@garyrussell i saw you were involved in the discussion, i'm just not familiar with the process here, how do can we get the pr reviewed?

garyrussell:
Someone just has to pick it up; I have been out sick the last couple of weeks; as such, I have a huge backlog so I may not get to this for a while.

PR #396, opened 2023-10-31
GH-395: make the retry context aware of the policies that have a maxi…

e-ivaldi:


src/main/java/org/springframework/retry/support/RetryTemplate.java
```diff
@@ -296,6 +296,10 @@ protected <T, E extends Throwable> T doExecute(RetryCallback<T, E> retryCallback
 				throw new TerminatedRetryException("Retry terminated abnormally by interceptor before first attempt");
 			}
 
+			if (!context.hasAttribute(RetryContext.MAX_ATTEMPTS)) {
```
e-ivaldi, line 299: potentially we could only set the value in the context when different from NO_MAXIMUM_ATTEMPTS_SET so that in the context it will be either  null or maxAttempts rather than NO_MAXIMUM_ATTEMPTS_SET or maxAttempts - probably the latter approach could be cleaner

artembilan (approved):
I have no objection on this.
I like the code as it is right now in the PR without any further changes.
Thanks

e-ivaldi:
> I have no objection on this. I like the code as it is right now in the PR without any further changes. Thanks

Glad to hear that @artembilan, thank you

artembilan (changes requested):
Add your name to the `@author` list of the affected classes, please.

e-ivaldi:
Thanks, done. Have added myself as an author to all classes I modified that already had an author list.

artembilan (changes requested):
Looks like there is a conflict in your branch against the latest `main`:
```
Conflicting files
src/test/java/org/springframework/retry/annotation/EnableRetryTests.java 
```
Please, consider to rebase, fix it and push.

Thanks

e-ivaldi:
Oh yes, I noticed immediately after pushing the update, rebased and fixed the conflicts.

PR #399, opened 2023-11-08
Retryable with exponential backoff not working with delayExpression

R-Zer0:


R-Zer0:
> Any chances to have this covered by some unit test? See `ExponentialBackOffPolicyTests` as a candidate for addition. Thanks

Added

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -289,15 +289,27 @@ public synchronized long getSleepAndIncrement() {
...
+			}
 		}
 
 		public double getMultiplier() {
 			return this.multiplierSupplier != null ? this.multiplierSupplier.get() : this.multiplier;
 		}
 
 		public long getInterval() {
-			return this.intervalSupplier != null ? this.intervalSupplier.get() : this.interval;
+			if (initialIntervalSupplier == null) {
```
artembilan, line 307: Need `this.` prefix for this property use

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -289,15 +289,27 @@ public synchronized long getSleepAndIncrement() {
...
+				return this.interval;
+			}
+			else {
+				return this.interval == DEFAULT_INITIAL_INTERVAL ? this.initialIntervalSupplier.get() : this.interval;
```
artembilan, line 311: Don't we need to call `intervalSupplier` every time?
Looks like with your change we do that only once.

artembilan (changes requested):

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
R-Zer0: Only first time when we need initial value, further iteration store multiplayer value in the `interval` variable [here ](https://github.com/spring-projects/spring-retry/blob/903640ec21c02a6278e1d9415edb59f3f1f5c0be/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java#L286)

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
R-Zer0: fixed

artembilan (changes requested):
The build against your PR has failed.
Would you mind double checking it locally running `./mvnw clean verify` ?

R-Zer0:
> The build against your PR has failed. Would you mind double checking it locally running `./mvnw clean verify` ?

I found discrepancy between annotation [default delay value 1000](https://github.com/spring-projects/spring-retry/blob/13879ceb32d3be7cff504f19e9b1a97d6acc58b4/src/main/java/org/springframework/retry/annotation/Backoff.java#L58) [here 1000 too](https://github.com/spring-projects/spring-retry/blob/13879ceb32d3be7cff504f19e9b1a97d6acc58b4/src/main/java/org/springframework/retry/annotation/Backoff.java#L58) but [here it is 100](https://github.com/spring-projects/spring-retry/blob/13879ceb32d3be7cff504f19e9b1a97d6acc58b4/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java#L55) I suspect this value also should be 1000, pushing this change. Let me know if my understanding correct and this change acceptable.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -49,10 +49,10 @@ public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<Exponenti
 	protected final Log logger = LogFactory.getLog(this.getClass());
 
 	/**
-	 * The default 'initialInterval' value - 100 millisecs. Coupled with the default
+	 * The default 'initialInterval' value - 1000 millisecs. Coupled with the default
 	 * 'multiplier' value this gives a useful initial spread of pauses for 1-5 retries.
 	 */
-	public static final long DEFAULT_INITIAL_INTERVAL = 100L;
+	public static final long DEFAULT_INITIAL_INTERVAL = 1000L;
```
artembilan, line 55: This is some kind of breaking change for the current point release, but I indeed see the reasoning behind consistency.
Plus it looks like the `@Backoff` annotation is always contributing exactly that `1000`  value.
So, we really can bite a bullet and let this pass through to the next release in a couple weeks.

artembilan (approved):
LGTM.

Let's hear yet from @garyrussell !

R-Zer0:
@artembilan pushed minor update to if else logic, to make it more readable

artembilan (approved):

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -289,15 +289,25 @@ public synchronized long getSleepAndIncrement() {
 		}
 
 		protected long getNextInterval() {
-			return (long) (this.interval * getMultiplier());
+			if (this.initialIntervalSupplier == null || this.interval != DEFAULT_INITIAL_INTERVAL) {
+				return (long) (this.interval * getMultiplier());
+			}
+			else {
+				return (long) (this.initialIntervalSupplier.get() * getMultiplier());
```
garyrussell, line 296: This is incorrect - the multiplier must be applied to the previous interval; this always applies it to the initial interval.

src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
```diff
@@ -94,6 +94,22 @@ public void testMultiBackOff() {
...
+		ExponentialBackOffPolicy strategy = new ExponentialBackOffPolicy();
+		long seed = 40;
+		double multiplier = 1.2;
+		strategy.initialIntervalSupplier(() -> 40L);
+		strategy.setMultiplier(multiplier);
+		strategy.setSleeper(sleeper);
+		BackOffContext context = strategy.start(null);
+		for (int x = 0; x < 5; x++) {
+			strategy.backOff(context);
+			assertThat(sleeper.getLastBackOff()).isEqualTo(seed);
```
garyrussell, line 108: We need to test that we still have exponential back offs (which we no longer have with the current change).

garyrussell (changes requested):

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
artembilan: Yes, that's why there is that logic in the `if`: `|| this.interval != DEFAULT_INITIAL_INTERVAL`.
So, the `initialIntervalSupplier` is called only once, for the first time. The next `interval` is really based on the previous one from now on.

R-Zer0: I've made this by purpose: during first iteration we should check if supplier provided or not, if provided first iteration should use value provided by `initialIntervalSupplier`(if present), further it will use `interval` value since it will be assigned to `interval` [here ](https://github.com/spring-projects/spring-retry/blob/27011f1bc787e3af9caca1369a5bc02a4c7c57e6/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java#L286)

src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
R-Zer0: Could you suggest any examaple here?

R-Zer0: Actually current test checking following, validated in debug.
| Last backoff | Seed |
|--------|--------|
| 40| 40 |
| 48| 48 |
| 57| 57 |
| 68| 68 |
| 81| 81 | 

I was thinking it is sufficient for current issue.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
garyrussell: But, if the supplier happens to return `DEFAULT_INITIAL_INTERVAL`, it will be called every time.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -289,15 +289,25 @@ public synchronized long getSleepAndIncrement() {
...
+			}
 		}
 
 		public double getMultiplier() {
 			return this.multiplierSupplier != null ? this.multiplierSupplier.get() : this.multiplier;
 		}
 
 		public long getInterval() {
-			return this.intervalSupplier != null ? this.intervalSupplier.get() : this.interval;
+			if (initialIntervalSupplier == null || this.interval != DEFAULT_INITIAL_INTERVAL) {
```
garyrussell, line 305: Why are we changing the logic to make the `interval` superior to the `intervalSupplier`?

Perhaps a description of exactly what you are trying to "fix" in the PR description would have been helpful.

garyrussell: Sorry, I am still missing something; with the existing code...

```
return this.intervalSupplier != null ? this.intervalSupplier.get() : this.interval;
```

...the supplier, if present, takes precedence over the `interval`, regardless of whether it is the default value or not.

Also, it is not at all clear to me why we are testing against `DEFAULT_INITIAL_INTERVAL` - this value could be reached at some point after the initial delay - e.g. if the initial interval returned by the supplier is `DEFAULT_INITIAL_INTERVAL / 2` and the `multiplier` is 2.0.

Perhaps you could provide a test case that fails with the existing code, it would help me understand what you are trying to fix.

R-Zer0: Sorry didn't understand line of code your referring, is it before my change?

You can try test I added in this PR it is failing without my change.
```
	@Test
	public void testMultiBackOffWithInitialDelaySupplier() {
		ExponentialBackOffPolicy strategy = new ExponentialBackOffPolicy();
		long seed = 40;
		double multiplier = 1.2;
		strategy.initialIntervalSupplier(() -> 40L);
		strategy.setMultiplier(multiplier);
		strategy.setSleeper(sleeper);
		BackOffContext context = strategy.start(null);
		for (int x = 0; x < 5; x++) {
			strategy.backOff(context);
			assertThat(sleeper.getLastBackOff()).isEqualTo(seed);
			seed *= multiplier;
		}
	}
```

artembilan: The problem, @garyrussell , that `getSleepAndIncrement()` when calling `this.interval = getNextInterval();` does not take into account `intervalSupplier`.
And this one we need to call there only once: for initial `interval` value since `this.interval` is not set in case of `intervalSupplier`.

garyrussell: Ok; I see the problem now, thanks; but I think this is a much simpler fix (and doesn't require any tests against the constant which, as I stated, may cause undesirable side effects).

```java
public long getInterval() {
	if (this.intervalSupplier != null) {
		this.interval = this.intervalSupplier.get();
		this.intervalSupplier = null;
	}
	return this.interval;
}
```

artembilan: ... and I guess we need to set its result into `this.interval` since `getNextInterval()` is going to use that one only.

garyrussell: Yes, 3rd line in the above code.

R-Zer0: updated according @garyrussell suggestion.

artembilan (changes requested):
Add your name to `@author` list, please, of the affected classes.

R-Zer0:
> Add your name to `@author` list, please, of the affected classes.

Requested update pushed

artembilan:
@R-Zer0 ,

PR #400, opened 2023-11-13
Add BackOffByExceptionTypePolicy to backoff by exception type

tony-clarke-amdocs:
This PR is an attempt to support a backoff policy based on the exception type thrown. It's similar to how we have a different retry policy based on the exception thrown. 

I've noticed that one of the serialization tests fails. This serialization requirement is not clear to me but perhaps it explains why such a backoff policy has not already been provided. If anyone could provide any ideas about how to support the serialization requirement that would be great.

### DIFF
--- a/src/test/java/org/springframework/retry/backoff/BackOffPolicySerializationTests.java
+++ b/src/test/java/org/springframework/retry/backoff/BackOffPolicySerializationTests.java
@@ -16,14 +16,13 @@
 
 package org.springframework.retry.backoff;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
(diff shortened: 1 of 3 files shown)


src/test/java/org/springframework/retry/backoff/BackOffPolicySerializationTests.java
```diff
@@ -16,14 +16,13 @@
 
 package org.springframework.retry.backoff;
 
-import java.util.ArrayList;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
manuelondina, line 19: this shouldn't be done

artembilan:

Issue #406, opened 2023-12-09
Issue with aspectjweaver  dependency used since version 2.0.0


staniakm:
There is an issue with delay when multiplierExpression is used. It starts since version 2.0.0. I've found that issue probably comes from aspectjweaver. 
I have simple spring service writen in kotlin and really simple test for it. Test will fail with spring-retry 2.0.0 up to 2.0.4 but will pass for version 1.3.4
After bumping aspectjweaver  from 1.9.20 to 1.9.20.1 building spring-retry locally problem was solved and test has passed

```@Service
class RetiedService() {
    @Retryable(IllegalArgumentException::class,
        backoff = Backoff(
            delayExpression = "\${setup.delay}",
            multiplierExpression = "\${setup.multiplier}",
            ),
        maxAttempts = 4)
    fun toBeRetried() {
        println ("${Instant.now()} Running method")
        willThrow()
    }

    private fun willThrow() {
        throw IllegalArgumentException("some exception")
    }
}
```

Properties
```
setup.delay=500
setup.multiplier=2.0
```
And test
```
@SpringBootTest
class Demo3ApplicationTests {

    @Autowired
    lateinit var service: RetiedService

    @Test
    fun contextLoads() {
        val time = measureTimeMillis {
            assertThrows<IllegalArgumentException> {
                service.toBeRetried()
            }
        }
        println(time)
        assert(time > 3500)
    }
}
```

artembilan:
So, if you confirm that updating to the latest AspectJ `1.9.20.1` fixes the problem, may  that is what enough to do in this project and the fix will be available in the next version we will plan to release this week?
If that a case, feel free to contribute such a fix.
Maybe your simple test would be great to have as an addition.

Thanks

Issue #397, opened 2023-11-01
Retryable with exponential backoff not working with delayExpression


mreichenbach-igniti:
Hi,

I tried to use the `@Retryable` annotation in spring boot with a configurable delay and exponential backoff. This is how I tried it:
```
@Retryable(
    maxAttemptsExpression = "3",
    backoff = @Backoff(delayExpression = "600", multiplier = 2, maxDelay = 10000)
)
```
The `maxAttemptsExpression` and `delayExpression` are actually filled with environment variables via `"${...}"`

**Expected**: exponential delay as described in the documentation

**Actual**: constant delay of 600ms

When I change it to `delay = 600` then it's working as expected. I checked it with spring-retry 2.0.4

Thanks for having a look!
Regards, Michael

R-Zer0:
Same with spring retry 2.0.3 for me

Noticed that in the `ExponentialBackOffContext` when we are calling `getSleepAndIncrement`
`sleep` value resolved from `intervalSupplier` if it is not null. But further when we resolving `getNextInterval`, we assigning it to the `interval` instead of  `intervalSupplier` and interval keep it initial value for the next iteration.

```
		public synchronized long getSleepAndIncrement() {
			long sleep = getInterval();
			long max = getMaxInterval();
			if (sleep > max) {
				sleep = max;
			}
			else {
				this.interval = getNextInterval();
			}
			return sleep;
		}
```

artembilan:
@mreichenbach-igniti , @R-Zer0 ,

if you see how that can be fixed, feel free to raise a respective Pull Request.

Thanks

R-Zer0:
@artembilan https://github.com/spring-projects/spring-retry/pull/399

PR #407, opened 2023-12-11
Upgrade to AspectJ 1.9.20.1

staniakm:


artembilan (changes requested):
Thanks for the update, but any chances that you can add a respective unit test which fails without that version bumping?

staniakm:
> Thanks for the update, but any chances that you can add a respective unit test which fails without that version bumping?

I'm looking for a way to unit test it but I don't see any tests that check if retry mechanism waits for some period of time before retry.

artembilan:
Well, you showed one in the issue: https://github.com/spring-projects/spring-retry/issues/406.
Then I see that we have `EnableRetryTests.ExpressionService` which provides a config similar to what you show in the issue.
Perhaps that's from where you can start unit testing it.

Would be great to see how it fails for you, so then we might be able to even just adjust existing test (and its configuration) to fail.
If it is only about Kotlin, then yeah, we don't have choice unless just accept the fix as is.

staniakm:
Nah, its also fails with java. Made the same test with spring java service. 

 Yeah, I have found this test `EnableRetryTests.ExpressionService`. Not yet sure how to calcualte elapsed time because now method execution finish really quick without consider delays defined by backoff.

artembilan:
I'm not sure how is the timing is relevant to the failure you are claiming with AspectJ.

Any explanations in behavior difference before and after upgrade?

staniakm:
I don't really know why aspectJ affect retry delay. But with version 2.0.* first retry wait delayExpresion time, but next one is not using multipliedExpression to calculate delay and still using delayExpresion. So, for example with delayExpression="500" and multiplier "2.0", delay between method execution will be 500 every time. With old verions 1.3+ or with upadted aspectJ, delay will use multiplier, so delays between next method executions will be 500, 1000, 2000...

I have found solution how to calculate time in unit tests using DummySleeper, but it looks like unit tests are not affected by aspectJ issu

staniakm:
Release notes for aspectJ https://htmlpreview.github.io/?https://github.com/eclipse/org.aspectj/blob/master/docs/dist/doc/README-1920.html
Ans issue rised in spring project https://github.com/spring-projects/spring-framework/issues/27761

artembilan:
Well, this one:
```
	@Retryable(backoff = @Backoff(delayExpression = "500", multiplierExpression = "2.0"),
			maxAttempts = 4)
	public void toBeRetried() {
		throw new IllegalArgumentException("some exception");
	}
```
fails for me even with AspectJ `1.9.20.1`:
```
2023-12-12T15:08:13.985-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Retry: count=0
2023-12-12T15:08:13.987-05:00 DEBUG 22384 --- [    Test worker] o.s.r.backoff.ExponentialBackOffPolicy   : Sleeping for 500
2023-12-12T15:08:14.502-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Checking for rethrow: count=1
2023-12-12T15:08:14.502-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Retry: count=1
2023-12-12T15:08:14.502-05:00 DEBUG 22384 --- [    Test worker] o.s.r.backoff.ExponentialBackOffPolicy   : Sleeping for 500
2023-12-12T15:08:15.011-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Checking for rethrow: count=2
2023-12-12T15:08:15.011-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Retry: count=2
2023-12-12T15:08:15.011-05:00 DEBUG 22384 --- [    Test worker] o.s.r.backoff.ExponentialBackOffPolicy   : Sleeping for 500
2023-12-12T15:08:15.512-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Checking for rethrow: count=3
2023-12-12T15:08:15.512-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Retry: count=3
2023-12-12T15:08:15.512-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Checking for rethrow: count=4
2023-12-12T15:08:15.513-05:00 DEBUG 22384 --- [    Test worker] o.s.retry.support.RetryTemplate          : Retry failed last attempt: count=4
```
but if I do this:
```
  implementation 'org.springframework.retry:spring-retry:2.0.5-SNAPSHOT'
```
Then all is good.
Apparently you are facing the problem fixed in this issue: https://github.com/spring-projects/spring-retry/issues/397.

So, I guess we can accept your PR as just a dependency upgrade and that's it.
Just because the problem has nothing to do with AspectJ and the real fix for `delayExpression` is already in.

We are planning to release `2.0.5` this Friday.

artembilan:
@staniakm ,

PR #424, opened 2024-02-16
Update README.md

### LINKED ISSUE #423: Wrong method recommended for build.gradle?
You recommend adding ` runtime('org.springframework.boot:spring-boot-starter-aop')` as a dependency to the project, but then I get _No candidates for method call runtime_. Did you mean
```
runtimeOnly 'org.springframework.boot:spring-boot-starter-aop'
```
?

banan314:
fixes #423 and also uses `retryFor` instead of deprecated `include` or `value`

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ public class Application {
 
 @Service
 class Service {
-    @Retryable(RemoteAccessException.class)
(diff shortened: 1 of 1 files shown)


artembilan:
@banan314 ,

Issue #427, opened 2024-04-27
@Retryable annotation retrying with ExponentialBackOff instead of ExponentialRandomBackOff when randomExpression provided


aftabshk:
I am using following versions:

Java - 17
Spring Retry - 2.0.5
Spring Boot - 3.2.5
Example code to reproduce issue: [spring-retry-issue](https://github.com/aftabshk/spring-retry-issue/tree/main)

When I am using the @Retryable annotation like following on any method:

`@Retryable(retryFor = {RuntimeException.class},
            maxAttempts = 10,
            backoff = @Backoff(delay = 2000,
                    multiplier = 2,
                    random = true))`

Everything works as expected. I get 10 retries exponentially, increasing and with **randomness** in time. But, when I use the **exact same values** but pass them via application.properties as follows:

`@Retryable(retryFor = {RuntimeException.class},
            maxAttemptsExpression = "${retry.max-attempts}",
            backoff = @Backoff(delayExpression = "${retry.delay}",
                    multiplierExpression = "${retry.multiplier}",
                    randomExpression = "${retry.random}"))`

Then I get 10 retries but only exponentially increasing and **no randomness** in time

You can see the example code here: [spring-retry-issue](https://github.com/aftabshk/spring-retry-issue/tree/main)

I have identified the issue in the code of the **spring-retry** library and raised a corresponding Pull Request for the same. 
Here it is: [Pull request 428](https://github.com/spring-projects/spring-retry/pull/428)
Please have a look and let me know if this works.

gaofengIt:
您好，我已经收到您的邮件，将尽快给您回复。

PR #428, opened 2024-04-27
GH-427: Add fix for the issue with multiplierExpression and randomExp…

aftabshk:
This PR fixes the issue explained here: [Issue number 427](https://github.com/spring-projects/spring-retry/issues/427).

Please comment if there is any feedback.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -453,7 +453,7 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		String randomExpression = (String) attrs.get("randomExpression");
 		Expression parsedRandomExp = null;
 
-		if (multiplier > 0 || multiplierExpression != null) {
+		if (multiplier > 0 || parsedMultExp != null) {
 			isRandom = backoff.random();
 			if (StringUtils.hasText(randomExpression)) {
 				parsedRandomExp = parse(randomExpression);
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -453,7 +453,7 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		String randomExpression = (String) attrs.get("randomExpression");
 		Expression parsedRandomExp = null;
 
-		if (getMultiplier(multiplier, parsedMultExp, stateless) > 0) {
+		if (multiplier > 0 || multiplierExpression != null) {
 			isRandom = backoff.random();
 			if (StringUtils.hasText(randomExpression)) {
 				parsedRandomExp = parse(randomExpression);
@@ -465,11 +465,6 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		}
 		return buildBackOff(min, parsedMinExp, max, parsedMaxExp, multiplier, parsedMultExp, isRandom, parsedRandomExp,
 				stateless);
-	}
-
-	private Double getMultiplier(Double multiplier, Expression parsedMultiplierExpression, Boolean stateless) {
-		return parsedMultiplierExpression != null ? evaluate(parsedMultiplierExpression, Double.class, stateless)
-				: multiplier;
 	}
 
 	private BackOffPolicy buildBackOff(long min, Expression minExp, long max, Expression maxExp, double multiplier,
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -76,6 +76,7 @@ import org.springframework.util.StringUtils;
  * @author Artem Bilan
  * @author Gary Russell
  * @author Roman Akentev
+ * @author Aftab Shaikh
  * @since 1.1
  */
 public class AnnotationAwareRetryOperationsInterceptor implements IntroductionInterceptor, BeanFactoryAware {
@@ -451,7 +452,8 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		boolean isRandom = false;
 		String randomExpression = (String) attrs.get("randomExpression");
 		Expression parsedRandomExp = null;
-		if (multiplier > 0) {
+
+		if (getMultiplier(multiplier, parsedMultExp, stateless) > 0) {
 			isRandom = backoff.random();
 			if (StringUtils.hasText(randomExpression)) {
 				parsedRandomExp = parse(randomExpression);
@@ -463,6 +465,11 @@ public class AnnotationAwareRetryOperationsInterceptor implements IntroductionIn
 		}
 		return buildBackOff(min, parsedMinExp, max, parsedMaxExp, multiplier, parsedMultExp, isRandom, parsedRandomExp,
 				stateless);
+	}
+
+	private Double getMultiplier(Double multiplier, Expression parsedMultiplierExpression, Boolean stateless) {
+		return parsedMultiplierExpression != null ? evaluate(parsedMultiplierExpression, Double.class, stateless)
+				: multiplier;
 	}
 
 	private BackOffPolicy buildBackOff(long min, Expression minExp, long max, Expression maxExp, double multiplier,
--- a/src/main/java/org/springframework/retry/annotation/Backoff.java
+++ b/src/main/java/org/springframework/retry/annotation/Backoff.java
@@ -130,9 +130,10 @@ public @interface Backoff {
 	 * Evaluates to a value. In the exponential case ({@link #multiplier()} &gt; 1.0) set
 	 * this to true to have the backoff delays randomized, so that the maximum delay is
 	 * multiplier times the previous delay and the distribution is uniform between the two
-	 * values. This expression is always evaluated during initialization. If the expression
-	 * returns true then {@link org.springframework.retry.backoff.ExponentialRandomBackOffPolicy}
-	 * is used else {@link org.springframework.retry.backoff.ExponentialBackOffPolicy} is used.
+	 * values. This expression is always evaluated during initialization. If the
+	 * expression returns true then
+	 * {@link org.springframework.retry.backoff.ExponentialRandomBackOffPolicy} is used
+	 * else {@link org.springframework.retry.backoff.ExponentialBackOffPolicy} is used.
 	 * @return the flag to signal randomization is required (default false)
 	 */
 	String randomExpression() default "";
--- a/src/main/java/org/springframework/retry/annotation/Backoff.java
+++ b/src/main/java/org/springframework/retry/annotation/Backoff.java
@@ -41,6 +41,7 @@ import org.springframework.retry.backoff.BackOffPolicy;
  *
  * @author Dave Syer
  * @author Gary Russell
+ * @author Aftab Shaikh
  * @since 1.1
  *
  */
@@ -130,7 +131,10 @@ public @interface Backoff {
 	 * this to true to have the backoff delays randomized, so that the maximum delay is
 	 * multiplier times the previous delay and the distribution is uniform between the two
 	 * values. Use {@code #{...}} for one-time evaluation during initialization, omit the
-	 * delimiters for evaluation at runtime.
+	 * delimiters for evaluation at runtime. This expression is evaluated at configuration
+	 * phase. If it returns true and ({@link #multiplier()} &gt; 1.0) then
+	 * {@link org.springframework.retry.backoff.ExponentialRandomBackOffPolicy} is created
+	 * else {@link org.springframework.retry.backoff.ExponentialBackOffPolicy} is created
 	 * @return the flag to signal randomization is required (default false)
 	 */
 	String randomExpression() default "";
--- a/src/main/java/org/springframework/retry/annotation/Backoff.java
+++ b/src/main/java/org/springframework/retry/annotation/Backoff.java
@@ -131,10 +131,10 @@ public @interface Backoff {
 	 * this to true to have the backoff delays randomized, so that the maximum delay is
 	 * multiplier times the previous delay and the distribution is uniform between the two
 	 * values. Use {@code #{...}} for one-time evaluation during initialization, omit the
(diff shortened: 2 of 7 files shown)


src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -451,7 +452,8 @@ private BackOffPolicy getBackoffPolicy(Backoff backoff, boolean stateless) {
 		boolean isRandom = false;
 		String randomExpression = (String) attrs.get("randomExpression");
 		Expression parsedRandomExp = null;
-		if (multiplier > 0) {
```
artembilan, line 454: I think the problem is exactly with this `if`.
See docs on those annotation attributes:
```
	/**
	 * Evaluates to a value used as a multiplier for generating the next delay for
	 * backoff; ; any value less than or equal to 1.0 is treated as 1.0, meaning a fixed
	 * delay. Overrides {@link #multiplier()}. Use {@code #{...}} for one-time evaluation
	 * during initialization, omit the delimiters for evaluation at runtime.
	 * @return a multiplier expression to use to calculate the next backoff delay (default
	 * 0 = ignored)
	 * @since 1.2
	 */
	String multiplierExpression() default "";
```

So, if you don't use SpEL template, then it is expected to be evaluated at runtime for each call, therefore that `multiplier` is never going to be `> 0` and we just don't determine anything for `random`.

src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
```diff
@@ -217,7 +218,7 @@ public BackOffPolicyBuilder randomSupplier(Supplier<Boolean> randomSupplier) {
 	public BackOffPolicy build() {
 		if (this.multiplier != null && this.multiplier > 0 || this.multiplierSupplier != null) {
 			ExponentialBackOffPolicy policy;
-			if (Boolean.TRUE.equals(this.random)) {
+			if (isRandom()) {
```
artembilan, line 221: I see that `randomSupplier` is ignored in this class and apparently it cannot be a runtime option anyway: we have to determine a `BackOffPolicy` instance during this `build()` call.
So, I find the fix for this class as good one, but probably `@Backoff(randomExpression)` Javadocs need to be fixed that this expression is always evaluated at configuration phase.

artembilan (changes requested):
Thank you for the contribution!
Let me know if my review makes sense for you!

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
aftabshk: Hi @artembilan , Thank you so much for your replies. 

The multiplierExpression was allowing values between 0 to 1 and was not treating it as 1. So if you provide a expression 0.5 then you will be able to see the retry time getting decreased. I added fix for that as well. 
You can see example of that issue being replicated here: [spring-retry-issue](https://github.com/aftabshk/spring-retry-issue)

Do you expect anything else on this?

src/main/java/org/springframework/retry/backoff/BackOffPolicyBuilder.java
aftabshk: @artembilan as suggested, javadoc has been fixed. Please have a look.

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -451,7 +452,8 @@ private BackOffPolicy getBackoffPolicy(Backoff backoff, boolean stateless) {
...
+
+		if (getMultiplier(multiplier, parsedMultExp, stateless) > 0) {
```
artembilan, line 456: This is still not correct: the `multiplier` might be calculated at runtime against each call.
That's exactly the point of those SpEL expressions.
I suggested before that this `random` is calculated unconditionally.

See the logic in the `BackOffPolicyBuilder`:
```
if (this.multiplier != null && this.multiplier > 0 || this.multiplierSupplier != null) {
```
So, if runtime expression is provided (read `multiplierSupplier`), then we don't care to what value it is going to be resolved.
We just cannot always resolve `parsedMultiplierExpression` at configuration phase since it might not be end-user intention.
Therefore that new `getMultiplier()` must be removed.

src/main/java/org/springframework/retry/annotation/Backoff.java
```diff
@@ -130,7 +131,10 @@
 	 * this to true to have the backoff delays randomized, so that the maximum delay is
 	 * multiplier times the previous delay and the distribution is uniform between the two
 	 * values. Use {@code #{...}} for one-time evaluation during initialization, omit the
-	 * delimiters for evaluation at runtime.
+	 * delimiters for evaluation at runtime. This expression is evaluated at configuration
+	 * phase. If it returns true and ({@link #multiplier()} &gt; 1.0) then
+	 * {@link org.springframework.retry.backoff.ExponentialRandomBackOffPolicy} is created
+	 * else {@link org.springframework.retry.backoff.ExponentialBackOffPolicy} is created
```
artembilan, line 137: The assumption about `multiplier` is not correct: it simply could be resolved at runtime as we discuss with you in other place.
So, consider to revise this one in exactly what we want to say: unlike other expressions, this one is always evaluated at configuration phase, so the framework can determine to build `ExponentialRandomBackOffPolicy` or `ExponentialBackOffPolicy`.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -172,7 +172,12 @@ public void initialIntervalSupplier(Supplier<Long> initialIntervalSupplier) {
 	 */
 	public void multiplierSupplier(Supplier<Double> multiplierSupplier) {
 		Assert.notNull(multiplierSupplier, "'multiplierSupplier' cannot be null");
-		this.multiplierSupplier = multiplierSupplier;
+		if (multiplierSupplier.get() > 1.0) {
```
artembilan, line 175: Same problem here: it is a point of `multiplierSupplier` to evaluate it at runtime for every call.

artembilan (changes requested):

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
aftabshk: Ok. Sorry for the confusion. Got your point now

aftabshk: Please check now

src/main/java/org/springframework/retry/annotation/Backoff.java
aftabshk: This is fixed. Please check

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
aftabshk: This is reverted now

artembilan: I have a convenient rule: if you agree with me, there is no need to comment `Fixed`, `Done` etc.
I'll see all the requested changes in the next commit you push.
This way I would receive only a single email from this PR knowing that everything requested is there.
It is totally fine to discuss if you are disagree or I am wrong.

Only exception is to notify explicitly if PR was rebased. In this case GitHub doesn't sent an email 🤷

src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
```diff
@@ -451,7 +452,8 @@ private BackOffPolicy getBackoffPolicy(Backoff backoff, boolean stateless) {
 		boolean isRandom = false;
 		String randomExpression = (String) attrs.get("randomExpression");
 		Expression parsedRandomExp = null;
-		if (multiplier > 0) {
+
+		if (multiplier > 0 || multiplierExpression != null) {
```
artembilan, line 456: Looks like it supposed to be a `parsedMultExp` - exactly the one which is intended for runtime evaluation.
If that is not template, then it is evaluated right now and then `multiplier` is set.

src/main/java/org/springframework/retry/annotation/Backoff.java
```diff
@@ -130,7 +131,10 @@
 	 * this to true to have the backoff delays randomized, so that the maximum delay is
 	 * multiplier times the previous delay and the distribution is uniform between the two
 	 * values. Use {@code #{...}} for one-time evaluation during initialization, omit the
-	 * delimiters for evaluation at runtime.
+	 * delimiters for evaluation at runtime. This expression is always evaluated at
+	 * configuration phase. If the expression returns true then
+	 * {@link org.springframework.retry.backoff.ExponentialRandomBackOffPolicy} is used
+	 * else {@link org.springframework.retry.backoff.ExponentialBackOffPolicy} is used.
```
artembilan, line 137: This is still not fixed. The `random` is always evaluated on configuration phase. We have to not its result to determine the target `BackOffPolicy`.

artembilan (changes requested):

src/main/java/org/springframework/retry/annotation/Backoff.java
aftabshk: If you look at the following code: the policy is being determined by the expression's evaluation. So the above statement makes sense right?
`if (isRandom()) {
				policy = new ExponentialRandomBackOffPolicy();
			}
			
private boolean isRandom() {
		return (this.randomSupplier != null && Boolean.TRUE.equals(this.randomSupplier.get()))
				|| Boolean.TRUE.equals(this.random);
	}`

artembilan: It is, but read this whole Javadoc again.
I worry about the sentence:

```
Use {@code #{...}} for one-time evaluation during initialization, omit the delimiters for evaluation at runtime
```
And that is exactly what I'm asking you to fix.
The `random` expression is never evaluated at runtime.
Instead it is evaluated at configuration phase unconditionally.

Thanks

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
```diff
@@ -109,4 +110,40 @@ public void shouldCreateExponentialRandomBackOff() {
...
+			.multiplier(2)
+			.randomSupplier(() -> true)
+			.sleeper(mockSleeper)
+			.build();
+		assertThat(ExponentialBackOffPolicy.class.isAssignableFrom(backOffPolicy.getClass())).isTrue();
+		ExponentialBackOffPolicy policy = (ExponentialBackOffPolicy) backOffPolicy;
+		assertThat(policy.getInitialInterval()).isEqualTo(100);
+		assertThat(policy.getMaxInterval()).isEqualTo(1000);
+		assertThat(policy.getMultiplier()).isEqualTo(2);
+		assertThat(new DirectFieldAccessor(policy).getPropertyValue("sleeper")).isEqualTo(mockSleeper);
```
artembilan, line 146: Why do we worry about `sleeper` in these new tests?
Looks like one of them should verify for `ExponentialRandomBackOffPolicy` and another for `ExponentialBackOffPolicy`.

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
```diff
@@ -109,4 +110,40 @@ public void shouldCreateExponentialRandomBackOff() {
...
+	}
+
+	@Test
+	public void shouldCreateExponentialBackOffWhenProvidedRandomSupplier() {
+		Sleeper mockSleeper = mock(Sleeper.class);
+		BackOffPolicy backOffPolicy = BackOffPolicyBuilder.newBuilder()
+			.delay(100)
+			.maxDelay(1000)
+			.multiplier(2)
+			.randomSupplier(() -> true)
```
artembilan, line 138: I don't understand what is difference in configuration for these tests.
Looks like both of them do `.randomSupplier(() -> true)`.
But how that happens that one of them is about `ExponentialRandomBackOffPolicy`, but another is not?

artembilan (changes requested):

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
aftabshk: This test is testing the build method. And inside build we are also setting the sleeper, so it should be tested in this test right?

artembilan: I see that `sleeper` is covered in other tests.
And does not look like it is required.
Therefore I'd prefer to have specific test concentrated on its purpose.
The point is that `new DirectFieldAccessor(policy).getPropertyValue("sleeper")` is a bit slow to have in every single unit test.

src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
```diff
@@ -109,4 +110,60 @@ public void shouldCreateExponentialRandomBackOff() {
...
 	}
 
+	@Test
+	public void shouldCreateExponentialRandomBackOffWhenProvidedRandomSupplier() {
+		Sleeper mockSleeper = mock(Sleeper.class);
+		BackOffPolicy backOffPolicy = BackOffPolicyBuilder.newBuilder()
+			.delay(10000)
+			.maxDelay(100000)
+			.multiplier(10)
+			.randomSupplier(() -> true)
```
artembilan, line 120: Can this also be like `random(true)`?
For consistency with the rest of properties you set here.

aftabshk: That test is already present. The purpose of this test is to test the scenario where random is provided via expression whereas other properties are values.

artembilan (changes requested):
The build has failed:
```
 Error:  Failed to execute goal io.spring.javaformat:spring-javaformat-maven-plugin:0.0.40:validate (default) on project spring-retry: Formatting violations found in the following files:
Error:   * /home/runner/work/spring-retry/spring-retry/src/main/java/org/springframework/retry/annotation/Backoff.java
Error:   * /home/runner/work/spring-retry/spring-retry/src/test/java/org/springframework/retry/backoff/BackOffPolicyBuilderTests.java
Error:  
Error:  Run `spring-javaformat:apply` to fix.
```
When you fix that and build is green, I'll merge.
Thanks

artembilan:
Has been merged.

Issue #441, opened 2024-06-01
Allow use of custom (binary) exception classifier in RestTemplateBuilder


morulay:
Some DBs hint with specific vendor code or `SQLState` if a DB request/transaction could be retried in case of exception, e.g. `exception instanceof SQLException sqlException && "12345".equals(sqlException.getErrorCode())` . `spring-retry` can handle this case either with:

- Utility class holding the predicate (classifier) + SPeL and `ExpressionRetryPolicy` - not very fast due to the expression evaluation
- Assembling manually `RestTemplate` with custom `RetryPolicy` - more error prone and lacks the convenience of `RetryTemplate.builder()`

Both solutions have drawbacks and mostly lacks the convenience of `RetryTemplate.builder()`. 

Do you think it is possible to allow the use of custom (binary) exception classifier in `RetryTemplate.builder()`. Here is an example to depict the idea:

```java
RetryTemplate retryTemplate = RetryTemplate.builder()
  .maxAtmepts(3)
  .customExceptionClassifier(classifiable -> classifiable instanceof SQLException sqlException
      && "retry_vendor_code".equals(sqlException.getErrorCode()))
  .build()
```

PR #442, opened 2024-06-01
Allow custom binary exception classifier with RestTemplateBuilder

morulay:
Proposed implementation for #441

### DIFF
--- a/src/main/java/org/springframework/retry/policy/BinaryExceptionClassifierRetryPolicy.java
+++ b/src/main/java/org/springframework/retry/policy/BinaryExceptionClassifierRetryPolicy.java
@@ -17,6 +17,7 @@
 package org.springframework.retry.policy;
 
 import org.springframework.classify.BinaryExceptionClassifier;
+import org.springframework.classify.Classifier;
 import org.springframework.retry.RetryContext;
 import org.springframework.retry.RetryPolicy;
 import org.springframework.retry.context.RetryContextSupport;
@@ -28,16 +29,15 @@ import org.springframework.retry.context.RetryContextSupport;
  *
  * @author Aleksandr Shamukov
  */
-@SuppressWarnings("serial")
 public class BinaryExceptionClassifierRetryPolicy implements RetryPolicy {
 
-	private final BinaryExceptionClassifier exceptionClassifier;
+	private final Classifier<Throwable, Boolean> exceptionClassifier;
 
-	public BinaryExceptionClassifierRetryPolicy(BinaryExceptionClassifier exceptionClassifier) {
+	public BinaryExceptionClassifierRetryPolicy(Classifier<Throwable, Boolean> exceptionClassifier) {
 		this.exceptionClassifier = exceptionClassifier;
 	}
 
-	public BinaryExceptionClassifier getExceptionClassifier() {
+	public Classifier<Throwable, Boolean> getExceptionClassifier() {
 		return exceptionClassifier;
 	}
 
--- a/src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
+++ b/src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
@@ -23,9 +23,9 @@ import org.springframework.retry.RetryPolicy;
 import org.springframework.retry.context.RetryContextSupport;
 
 /**
- * A policy, that is based on {@link Predicate<Throwable>}. Usually, binary
- * classification is enough for retry purposes. If you need more flexible classification,
- * use {@link ExceptionClassifierRetryPolicy}.
+ * A policy, that is based on {@link Predicate<Throwable>}. Usually, binary classification
+ * is enough for retry purposes. If you need more flexible classification, use
+ * {@link ExceptionClassifierRetryPolicy}.
  *
  * @author Morulai Planinski
  */
--- a/src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
+++ b/src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
@@ -16,35 +16,35 @@
 
 package org.springframework.retry.policy;
 
-import org.springframework.classify.BinaryExceptionClassifier;
+import java.util.function.Predicate;
+
 import org.springframework.retry.RetryContext;
 import org.springframework.retry.RetryPolicy;
 import org.springframework.retry.context.RetryContextSupport;
 
 /**
- * A policy, that is based on {@link BinaryExceptionClassifier}. Usually, binary
+ * A policy, that is based on {@link Predicate<Throwable>}. Usually, binary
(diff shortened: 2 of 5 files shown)


src/main/java/org/springframework/retry/policy/BinaryExceptionClassifierRetryPolicy.java
```diff
@@ -28,16 +29,15 @@
...
  * @author Aleksandr Shamukov
  */
-@SuppressWarnings("serial")
 public class BinaryExceptionClassifierRetryPolicy implements RetryPolicy {
 
-	private final BinaryExceptionClassifier exceptionClassifier;
+	private final Classifier<Throwable, Boolean> exceptionClassifier;
 
-	public BinaryExceptionClassifierRetryPolicy(BinaryExceptionClassifier exceptionClassifier) {
+	public BinaryExceptionClassifierRetryPolicy(Classifier<Throwable, Boolean> exceptionClassifier) {
```
artembilan, line 36: While I agree with you, this is a breaking change for the current `2.0.x` version.
There is already the `ExceptionClassifierRetryPolicy` which supports any arbitrary:
```
	/**
	 * Setter for an exception classifier. The classifier is responsible for translating
	 * exceptions to concrete retry policies. Either this property or the policy map
	 * should be used, but not both.
	 * @param exceptionClassifier ExceptionClassifier to use
	 */
	public void setExceptionClassifier(Classifier<Throwable, RetryPolicy> exceptionClassifier) {
```

So, how about to use that one instead in the `RetryTemplateBuilder` when we provide such a `Predicate` you are talking about?

artembilan (changes requested):

src/main/java/org/springframework/retry/policy/BinaryExceptionClassifierRetryPolicy.java
morulay: I do agree `ExceptionClassifierRetryPolicy` is mighty option, but at the cost of complexity. Simple code like this:
```java
        return RetryTemplate.builder()
            .maxAttempts(3)
            .fixedBackoff(200)
            .retryOn(throwable -> throwable instanceof SQLException sqlException
                && sqlException.getErrorCode() == 256)
            .build();
```
becomes 20+ lines of code making it hard to maintain and error prone :-(

From my point of view, the constructor you address is backward compatible, because you can still use `BinaryExceptionClassifier` instance as an argument. It implements `Classifier<Throwable, Boolean>`

artembilan: But you have modified getter as well, which is not binary compatible.
Plus I didn't mean to reject your suggestion for the `RetryTemplateBuilder` API.
I just suggested to use that `ExceptionClassifierRetryPolicy` in the `build()` when simple `Predicate` is provided.
This way we definitely won't break any API and have a nice addition to the target functionality.

morulay: You're right about the getter. About ExceptionClassifierRetryPolicy it is my fault. I didn't get your idea. I'll try to update the PR following your advice.

morulay: After few experiments with ExceptionClassifierRetryPolicy I decided to introduce new PredicateRetryPolicy to produce more simple code. I'll be happy to have your opinion?

src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.retry.RetryContext;
+import org.springframework.retry.RetryPolicy;
+import org.springframework.retry.context.RetryContextSupport;
+
+/**
+ * A policy, that is based on {@link Predicate<Throwable>}. Usually, binary classification
+ * is enough for retry purposes. If you need more flexible classification, use
+ * {@link ExceptionClassifierRetryPolicy}.
+ *
+ * @author Morulai Planinski
```
artembilan, line 30: Yeah... Now I see your point.
This really should be much straightforward than trying to adapt `ExceptionClassifierRetryPolicy` or `ExpressionRetryPolicy`.

Please, add `@since 2.0.7`

src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
```diff
@@ -0,0 +1,65 @@
...
+public class PredicateRetryPolicy implements RetryPolicy {
+
+	private final Predicate<Throwable> predicate;
+
+	public PredicateRetryPolicy(Predicate<Throwable> predicate) {
+		this.predicate = predicate;
+	}
+
+	public Predicate<Throwable> getPredicate() {
+		return predicate;
```
artembilan, line 41: I don't see a reason in this getter.
I even don't see it in other classes for similar props, but the cat is out of the bag over there to complain.

src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
```diff
@@ -0,0 +1,65 @@
...
+ * {@link ExceptionClassifierRetryPolicy}.
+ *
+ * @author Morulai Planinski
+ */
+public class PredicateRetryPolicy implements RetryPolicy {
+
+	private final Predicate<Throwable> predicate;
+
+	public PredicateRetryPolicy(Predicate<Throwable> predicate) {
+		this.predicate = predicate;
```
artembilan, line 37: `Assert.notNull(predicate, "blah-blah")`

src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java
```diff
@@ -462,6 +466,26 @@ public RetryTemplateBuilder notRetryOn(List<Class<? extends Throwable>> throwabl
 		return this;
 	}
 
+	/**
+	 * Sets {@link Predicate<Throwable>} that decides if the exception causes a retry.
```
artembilan, line 470: Method Javadocs must be imperative, like a command.

artembilan (changes requested):
Please, add your name to the `@author` list of all the affected classes.

artembilan (approved):
OK. I'm going ahead and merging this with those minor cleanups.

artembilan:
Merged as https://github.com/spring-projects/spring-retry/commit/5b69d8059d1cf68dedd6d3b4eba23cab3f861e02.

src/main/java/org/springframework/retry/policy/PredicateRetryPolicy.java
morulay: I see there is the same in `BinaryExceptionClassifierRetryPolicy.getExceptionClassifier()` used solely for the purpose of testing. The one that made my first proposal backward incompatible

Issue #458, opened 2024-07-29
Built-in support for Micrometer metrics publishing


szpak:
It would be useful to provide Micrometer metrics related to spring-retry functionality (e.g. number of successful/failed retries (per retry/label) or retry execution times). Preferably, it could be done in the declarative way (no extra code/changes in the end project) - the feature is enabled and/or the extra module is added.

artembilan:
Thank you for the request!

If you have anything in mind how that supposed to be implemented on the `RetryTemplate`, I'll be happy to review.

Although from our experience an unconditional instrumentation is full of errors: no one said that we would like to instrument every single `RetryTemplate` instance in our application.
This could be like a high-level API (`setEnableMetrics()`) and respective `@Retriable` attribute.

We do have some stuff in the `org.springframework.retry.stats` with a `StatisticsListener` implementation.
Might be used as an inspiration (or even reused).
Perhaps this also can be implemented as a `RetryListener` contract to have a Micrometer as an `optional` dependency.

artembilan:
Looks like this project has some idea how to implement it: https://github.com/findinpath/spring-retry-metrics.

I still think that this can be implemented as an injectable `RetryListener`.
The auto-configuration for this feature is out of this project scope.
This is just library with an API without too much opinion on the default options and classpath decisions.

artembilan:
I wonder if `@Timed` feature from Micrometer does that trick already alongside with the `@Retryable`: https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_the_timed_annotation.
However I agree that it won't collect the number of attempts...

szpak:
I didn't have anything particular in mind reporting that idea. I've just thought, it could be useful to that those statistics/metrics available and Micrometer is an natural continuation.

The separate - Micrometer aware - module with injectable listener (and probably switchable auto configuration) sounds like a good idea. When added as a dependency (and enable with auto configuration), it could "observe" all the retryable methods, unless annotated with some dedicated annotation (`@NotObservlable` or something). Or as you suggested, to have some other way to point which methods/beans should be "observed" via the API.

There is a "new" [Micrometer observation API](https://softwaremill.com/new-micrometer-observation-api-with-spring-boot-3/) and maybe it could be leveraged (however, I haven't had a chance to play with it so far).

Btw, maybe @findinpath has any thoughts after the original implementation in his project (and encountered challenges), years ago.

artembilan:
[...]

PR #459, opened 2024-07-31
GH-458: Introduce `MetricsRetryListener`

artembilan:
Fixes: #458

### DIFF
--- a/src/main/java/org/springframework/retry/support/MetricsRetryListener.java
+++ b/src/main/java/org/springframework/retry/support/MetricsRetryListener.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.retry.support;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Function;
+
+import io.micrometer.core.instrument.Meter;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
+import io.micrometer.core.instrument.Timer;
+
+import org.springframework.retry.RetryCallback;
+import org.springframework.retry.RetryContext;
+import org.springframework.retry.RetryListener;
+import org.springframework.util.Assert;
+
+/**
+ * The {@link RetryListener} implementation for Micrometer {@link Timer}s around retry
+ * operations.
+ * <p>
+ * The {@link Timer#start} is called from the {@link #open(RetryContext, RetryCallback)}
+ * and stopped in the {@link #close(RetryContext, RetryCallback, Throwable)}. This
+ * {@link Timer.Sample} is associated with the provided {@link RetryCallback} to make this
+ * {@link MetricsRetryListener} instance reusable for many retry operation.
+ * <p>
+ * The registered {@value #TIMER_NAME} {@link Timer} has these tags by default:
+ * <ul>
+ * <li>{@code name} - {@link RetryCallback#getLabel()}</li>
+ * <li>{@code retry.count} - the number of attempts - 1; essentially the successful first call means no counts</li>
+ * <li>{@code exception} - the thrown back to the caller (after all the retry attempts) exception class name</li>
+ * </ul>
+ * <p>
+ * The {@link #setCustomTags(Map)} and {@link #setCustomTagsProvider(Function)} can be used to further customize
+ * tags on the timers.
+ *
+ * @author Artem Bilan
+ * @since 2.0.8
+ */
+public class MetricsRetryListener implements RetryListener {
+
+	public static final String TIMER_NAME = "spring.retry";
+
+	private final MeterRegistry meterRegistry;
+
+	private final Map<RetryCallback<?, ?>, Timer.Sample> retryContextToSample = new HashMap<>();
+
+	private final Map<String, String> customTags = new HashMap<>();
+
+	private final Meter.MeterProvider<Timer> retryMeterProvider;
+
+	private Function<RetryContext, Iterable<Tag>> customTagsProvider = retryContext -> Tags.empty();
+
+	/**
+	 * Construct an instance based on the provided {@link MeterRegistry}.
+	 * @param meterRegistry the {@link MeterRegistry} to use for timers.
+	 */
+	public MetricsRetryListener(MeterRegistry meterRegistry) {
+		Assert.notNull(meterRegistry, "'meterRegistry' must not be null");
+		this.meterRegistry = meterRegistry;
+		this.retryMeterProvider = Timer.builder(TIMER_NAME)
+			.description("Metrics for Spring RetryTemplate")
+			.withRegistry(this.meterRegistry);
+	}
+
+	/**
+	 * Supply a map of tags which are going to be used for all the timers managed by this
+	 * listener.
+	 * @param customTags the map of additional tags for all the timers.
+	 */
+	public void setCustomTags(Map<String, String> customTags) {
+		Assert.notNull(customTags, "'customTags' must not be null");
+		this.customTags.putAll(customTags);
+	}
+
+	/**
+	 * Supply a {@link Function} to build additional tags for all the timers based on the
+	 * {@link RetryContext}.
+	 * @param customTagsProvider the {@link Function} for additional tags with a
+	 * {@link RetryContext} scope.
+	 */
+	public void setCustomTagsProvider(Function<RetryContext, Iterable<Tag>> customTagsProvider) {
+		Assert.notNull(customTagsProvider, "'customTagsProvider' must not be null");
+		this.customTagsProvider = customTagsProvider;
+	}
+
+	@Override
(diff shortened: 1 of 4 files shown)


src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+
+	/**
+	 * Supply a map of tags which are going to be used for all the timers managed by this
+	 * listener.
+	 * @param customTags the map of additional tags for all the timers.
+	 */
+	public void setCustomTags(Map<String, String> customTags) {
+		Assert.notNull(customTags, "'customTags' must not be null");
+		this.customTags.putAll(customTags);
+	}
```
jonatan-ivanov, line 92: I would do this in the constructor instead, also If you use `Tags` instead of `Map` for `customTags` above, you can do this (`Tags` implements `Iterable<Tag>` ):
```suggestion
	public void setCustomTags(Iterable<Tag> customTags) {
		Assert.notNull(customTags, "'customTags' must not be null");
		this.customTags.and(customTags);
	}
```

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+			Throwable throwable) {
+
+		Timer.Sample sample = this.retryContextToSample.remove(callback);
+
+		Assert.state(sample != null,
+				() -> String.format("No 'Timer.Sample' registered for '%s'. Was the 'open()' called?", context));
+
+		Tags retryTags = Tags.of("name", callback.getLabel())
+			.and("retry.count", "" + context.getRetryCount())
+			.and(this.customTags.entrySet().stream().map((entry) -> Tag.of(entry.getKey(), entry.getValue())).toList())
```
jonatan-ivanov, line 122: If you use `Tags` instead of `Map` for `customTags` above, you can do this:
```suggestion
			.and(this.customTags)
```

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+		this.customTags.putAll(customTags);
+	}
+
+	/**
+	 * Supply a {@link Function} to build additional tags for all the timers based on the
+	 * {@link RetryContext}.
+	 * @param customTagsProvider the {@link Function} for additional tags with a
+	 * {@link RetryContext} scope.
+	 */
+	public void setCustomTagsProvider(Function<RetryContext, Iterable<Tag>> customTagsProvider) {
```
jonatan-ivanov, line 100: I think I would do this in a ctor.

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+
+	public static final String TIMER_NAME = "spring.retry";
+
+	private final MeterRegistry meterRegistry;
+
+	private final Map<RetryCallback<?, ?>, Timer.Sample> retryContextToSample = new HashMap<>();
+
+	private final Map<String, String> customTags = new HashMap<>();
+
+	private final Meter.MeterProvider<Timer> retryMeterProvider;
```
jonatan-ivanov, line 68: Fyi: this is a fairly new feature, if you use it, this will prevent users to use older versions of Micrometer. I don't think this is a big issue though, just wanted to give you a heads up. (Also `MeterProvider` might be more performant than using the builder.)

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	@Override
+	public <T, E extends Throwable> void close(RetryContext context, RetryCallback<T, E> callback,
+			Throwable throwable) {
+
+		Timer.Sample sample = this.retryContextToSample.remove(callback);
+
+		Assert.state(sample != null,
+				() -> String.format("No 'Timer.Sample' registered for '%s'. Was the 'open()' called?", context));
```
jonatan-ivanov, line 118: I like that this is checked especially since if `open` was called but the sample is not found here, it is a potential memory leak (if it wasn't removed by something else).

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+
+		Tags retryTags = Tags.of("name", callback.getLabel())
+			.and("retry.count", "" + context.getRetryCount())
+			.and(this.customTags.entrySet().stream().map((entry) -> Tag.of(entry.getKey(), entry.getValue())).toList())
+			.and(this.customTagsProvider.apply(context));
+
+		if (throwable != null) {
+			retryTags = retryTags.and("exception", throwable.getClass().getSimpleName());
+		}
```
jonatan-ivanov, line 127: This condition can make things tricky: it will mean that you can end up with time-series that has the same name but different set of tags (exception vs. no exception):
```
spring.retry{name=test, retry.count=0} ...
spring.retry{name=test, retry.count=0, exception=BOOM!} ...
```
This is usually discouraged but more than that, for the Prometheus Java client this is invalid and the metrics will be dropped. You can avoid this by adding a default value to the `exception` tag, for example `none` (MVC, WebFlux, etc. are doing this):
```
spring.retry{name=test, retry.count=0, exception=none} ...
spring.retry{name=test, retry.count=0, exception=BOOM!} ...
```
You can do something like this:
```java
private String exceptionTagValue(Throwable throwable) {
	if (throwable != null) {
		return throwable.getClass().getSimpleName();
	}
	else {
		return "none";
	}
}
```
and in `retryTags` simply:
```java
Tags retryTags = Tags.of("name", callback.getLabel())
	.and("retry.count", "" + context.getRetryCount())
	.and("exception", exceptionTagValue(throwable))
```

szpak:
Thanks @artembilan for your effort! I've just given it a try from the functionality point of view and after defining `MetricsRetryListener` as a bean, it just worked:

```
spring_retry_seconds_max{application="x",name="process",retry_count="0",} 0.005874044
spring_retry_seconds_max{application="x",name="process",retry_count="2",} 2.28533499
spring_retry_seconds_count{application="x",name="process",retry_count="0",} 2.0
spring_retry_seconds_count{application="x",name="process",retry_count="2",} 1.0
spring_retry_seconds_sum{application="x",name="process",retry_count="0",} 0.008557336
spring_retry_seconds_sum{application="x",name="process",retry_count="2",} 2.28533499
```

I will think over, if we would need anything else here, but for now it provides what we need.

szpak:
Btw, "seconds" looks strange at the first look in the context of the counter ;-).

artembilan:
> "seconds" looks strange at the first look in the context of the counter

Because it is not a `counter`, but `timer`. So, it gathers not only a number of call, but also some time measurements for those calls.
See Micrometer docs: https://docs.micrometer.io/micrometer/reference/concepts/timers.html

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+ * @author Artem Bilan
+ * @since 2.0.8
+ */
+public class MetricsRetryListener implements RetryListener {
+
+	public static final String TIMER_NAME = "spring.retry";
+
+	private final MeterRegistry meterRegistry;
+
+	private final Map<RetryCallback<?, ?>, Timer.Sample> retryContextToSample = new HashMap<>();
```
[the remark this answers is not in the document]
artembilan: No need to worry about concurrency: same retry context cannot be used from different threads in parallel.
However `IdentityHashMap` is a good suggestion.
I had the map before based on the `RetryContext`, but that one is modified in between `open` and `close`, so its hash is volatile.
But I believe an identity is a good compromise to not rely on the custom unexpected `RetryCallback` implementation.

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
```diff
@@ -0,0 +1,132 @@
...
+
+	private final Map<String, String> customTags = new HashMap<>();
```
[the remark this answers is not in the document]
artembilan: Yeah... Thought about an XML config support where use builder pattern is hard enough.
Plus `Map` API is familiar for users.
But yes, `Tags` is good enough to expose instead for consistency.

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
artembilan: I'm not sure what you mean with constructor, but since this is optional feature, then our style is to have it as setter.
But I agree that these custom tags can go to the `retryMeterProvider` immediately from here instead of later on in the `close`.

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
artembilan: Same reason about ctor here as well: it is optional.

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
artembilan: Now I know why I see `exception=none` in `spring-kafka` and `spring-integration`😄

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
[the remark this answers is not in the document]
jonatan-ivanov: You can also keep `Tags` as a `private` field and the ctor/setter parameter can be a `Map` (in addition `Tags` can be there too as an overload).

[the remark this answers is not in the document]
artembilan: Right. But since `Map` and `Tags` are same matter from end-user perspective, I'd prefer to not give them a paradox of choice.
If there no more feedback, I'll go ahead and incorporate what we have agreed upon so far.
Thanks

src/main/java/org/springframework/retry/support/MetricsRetryListener.java
artembilan: Well, no: the `MeterProvider` is not a builder and all of its methods returns for us a result `Meter`. 
So, taking that back. 
Now I see why you have insisted on moving it to the ctor! 😄

PR #465, opened 2024-08-31
Fix backOff API in UniformRandomBackOffPolicy

LokeshAlamuri:
When maxBackOffPeriod is less than minBackOffPeriod, delta is taken taken as zero in UniformRandomBackOffPolicy backOff method.

### DIFF
--- a/src/main/java/org/springframework/retry/backoff/UniformRandomBackOffPolicy.java
+++ b/src/main/java/org/springframework/retry/backoff/UniformRandomBackOffPolicy.java
@@ -134,7 +134,7 @@ public class UniformRandomBackOffPolicy extends StatelessBackOffPolicy
 	protected void doBackOff() throws BackOffInterruptedException {
 		try {
 			Long min = this.minBackOffPeriod.get();
-			long delta = this.maxBackOffPeriod.get() == this.minBackOffPeriod.get() ? 0
+			long delta = this.maxBackOffPeriod.get() <= this.minBackOffPeriod.get() ? 0
 					: this.random.nextInt((int) (this.maxBackOffPeriod.get() - min));
 			this.sleeper.sleep(min + delta);
 		}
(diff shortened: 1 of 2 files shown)


src/main/java/org/springframework/retry/backoff/UniformRandomBackOffPolicy.java
```diff
@@ -134,7 +134,7 @@ public long getMaxBackOffPeriod() {
 	protected void doBackOff() throws BackOffInterruptedException {
 		try {
 			Long min = this.minBackOffPeriod.get();
-			long delta = this.maxBackOffPeriod.get() == this.minBackOffPeriod.get() ? 0
+			long delta = this.maxBackOffPeriod.get() <= this.minBackOffPeriod.get() ? 0
 					: this.random.nextInt((int) (this.maxBackOffPeriod.get() - min));
```
artembilan, line 138: I think we need to extract both `Supplier.get()` into local variables and use them afterwards.
I think the logical way is possibility to return different values for every `get()` call.
So, it is OK to have them different for each `backOff()` call, but not within a single one.
Otherwise it may lead to unexpected  behavior.
I mean we have that `Long min = this.minBackOffPeriod.get();` variable.
So, let's use that in the following code!
Plus let's extract `Long max = this.maxBackOffPeriod.get();`

artembilan (changes requested):

LokeshAlamuri:
I have identified one more issue in `UniformRandomBackOffPolicy`. Please let me know if i have to raise another issue or should fix along with PR itself.

Issue: With UniformRandomBackOffPolicy there are no chances of getting `maxBackOffPeriod` as per the configuration. Maximum possible value is `maxBackOffPeriod -1`.

Reason:
In doBackOff please find the following code.

long delta = max <= min ? 0 : **_this.random.nextInt((int) (max - min))_**;

Random nextInt upper bound value is always less than the provided argument. So, backoffValue computated as `min + delta` will never touches `maxBackOffPeriod`.

Possible Fix (need verify one more time on this):
long delta = max <= min ? 0 : this.random.nextInt((int) (max - min + 1));

artembilan:
According to its JavaDocs:
```
uniformly distributed int value between 0 (inclusive) and the specified value (exclusive)
```
So, your observation is correct.
However we never advertised that `maxBackOffPeriod` is an inclusive max value.
Looks like the logic is to get delta somewhere between `0` and `max` and that is aligned with whatever `nextInt()` does.
Therefore, what you suggest is a behavior change where normal impression is to follow whatever `nextInt()` does.
At least that is what makes sense for me.

PR #472, opened 2024-09-23
Add default hashCode implementation to RetryConfiguration.AnnotationClassOrMethodPointcut

NorsaG:
This pull request introduces a default implementation of `hashCode` for `RetryConfiguration.AnnotationClassOrMethodPointcut`.

I discovered that using Spring Cloud with configuration refreshing (specifically the `@RefreshScope` annotation and `ConfigServicePropertySourceLocator` as part of spring-cloud-config) can lead to a "java.lang.OutOfMemoryError: Metaspace". This occurs because the `ConfigServicePropertySourceLocator` is re-initialized with every refresh. Due to the absence of a `hashCode` implementation, proxies for this class are regenerated each time, which causes them to be stored in the `org.springframework.cglib.core.internal.LoadingCache` (with `RetryConfiguration.AnnotationClassOrMethodPointcut` being part of the composite key for this object). 
This results in an excessive accumulation of generated proxies in memory, ultimately leading to metaspace overflow.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
+++ b/src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 
@@ -68,6 +69,7 @@ import org.springframework.util.ReflectionUtils;
  * @author Markus Heiden
  * @author Gary Russell
  * @author Yanming Zhou
+ * @author Evgeny Lazarev
  * @since 1.1
  *
  */
@@ -240,6 +242,11 @@ public class RetryConfiguration extends AbstractPointcutAdvisor
 			}
 			AnnotationClassOrMethodPointcut otherAdvisor = (AnnotationClassOrMethodPointcut) other;
 			return ObjectUtils.nullSafeEquals(this.methodResolver, otherAdvisor.methodResolver);
+		}
+
+		@Override
+		public int hashCode() {
+			return Objects.hash(this.methodResolver);
 		}
 
 	}


src/main/java/org/springframework/retry/annotation/RetryConfiguration.java
```diff
@@ -242,6 +243,11 @@ public boolean equals(Object other) {
 			return ObjectUtils.nullSafeEquals(this.methodResolver, otherAdvisor.methodResolver);
 		}
 
+		@Override
+		public int hashCode() {
+			return Objects.hash(methodResolver);
```
artembilan, line 248: Please, add `this.` prefix.

artembilan (changes requested):
Update Copyright of the changed class to actual year.
Add you name to the `@author` list.

Thanks

NorsaG:
Fit the PR according to the comment.

artembilan:
@NorsaG ,

NorsaG:
@artembilan, I’ve finally published an article with a detailed explanation of a real-world case of this issue: https://shorturl.at/HVa0Y
I think it could be interesting

artembilan:
Hi @NorsaG !

The article is great!
Sorry to cause so much pain and great to see how you have figured out everything.

Would you mind from your experience point me where exactly that `RetryConfiguration.AnnotationClassOrMethodPointcut` is used a part of cache key.

Thank you!

NorsaG:
@artembilan Yes, of course.

I've tried to reproduce the same problem again. It was a bit tricky due to the complexity of our project and the ongoing component upgrades. 

In Picture 1, you can observe the LoadingCache object at the moment of retrieving an object from the cache (I paused at a debug point after several configuration updates from the config server).


In Picture 2, nothing is found in the cache. 


In Picture 3, I have highlighted the path to the pointcut.  


(I'm not entirely sure, but I highlighted the "methodMatcherKey" string because it is unique for all EnhancerKeys in the cache.)

artembilan:
Thanks. I see now.
So, the path is like this: 
1. `org.springframework.aop.framework.AdvisedSupport`:
```
private List<Advisor> advisorKey = this.advisors;
```
Where one of them is our `RetryConfiguration`.
2. `org.springframework.aop.framework.CglibAopProxy.ProxyCallbackFilter`:
```
public int hashCode() {
			return this.advised.getAdvisorKey().hashCode();
		}
```
3. `org.springframework.cglib.proxy.Enhancer.createHelper()`:
```
		Object key = new EnhancerKey((superclass != null ? superclass.getName() : null),
				(interfaces != null ? Arrays.asList(ReflectUtils.getNames(interfaces)) : null),
				(filter == ALL_ZERO ? null : new WeakCacheKey<>(filter)),
				Arrays.asList(callbackTypes),
				useFactory,
				interceptDuringConstruction,
				serialVersionUID);
```

So, apparently the problem has gone with that our `hashCode()` fix just because the `RetryConfiguration` uses a default implementation based on its properties.

I see how this was hard to catch and even harder to determine what simple fix has to be.

Thank you again!

I will look now to other `Advisor`  implementations for similar problem.

NorsaG:
That was a really interesting issue! It was a bit of a pain to track down the cause since the production service wasn’t accessible, but I’m happy I could help out

artembilan:
I was wrong with that sentence:

> So, apparently the problem has gone with that our `hashCode()` fix just because the `RetryConfiguration` uses a default implementation based on its properties.

It is not about properties, but object identity by default: https://varoa.net/jvm/java/openjdk/biased-locking/2017/01/30/hashCode.html.

And you were right, the `AdvisorKeyEntry` is the root of the logic:
```
this.methodMatcherKey = pointcut.getMethodMatcher().toString();
```
That `getMethodMatcher()` is like this for our `RetryConfiguration.AnnotationClassOrMethodPointcut`:
```
	public final MethodMatcher getMethodMatcher() {
		return this;
	}
```

and that default `toString()` implementation is:
```
    public String toString() {
        return getClass().getName() + "@" + Integer.toHexString(hashCode());
    }
```
So, that's where we got that discrepancy because default `hashCode()` is an identity of the new object after refresh scope.

NorsaG:
I read the article above, but in my opinion, the author described a some different set of issues.

According to the official API note:

> It is generally necessary to override the [hashCode](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#hashCode()) method whenever this method is overridden, so as to maintain the general contract for the hashCode method, which states that equal objects must have equal hash codes.

[Source: Object.equals() documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Object.html#equals(java.lang.Object))

This was a significant problem in our case.

Another point is the `toString()` method. While debugging the latest version of the code, I noticed it, specifically when I saw the default `toString()` output for `methodMatcherKey`.
It might be a good idea to override this method, especially considering its usage here:
`this.methodMatcherKey = pointcut.getMethodMatcher().toString();`

artembilan:
I think `toString()` would be that unnecessary extra.
The main point that we have missed to implement properly `equals()` and `hashCode()`.
Nothing more to worry about!

Issue #480, opened 2024-12-14
Consider reverting change to revision attribute in pom


snicoll:
Please consider reverting https://github.com/spring-projects/spring-retry/issues/474 and everything that it caused. The purpose of the GitHub workflow is to make sure that things as much aligned as possible and each deviation from that makes Spring Retry special for no good reason.

The maven flatten plugin can easily be introduced to publish a POM that's more clean and without unnecessary information. That would be a plus as well.

artembilan:
Hey, Stephane!

Thank you for feedback!

So, my logic was like this:
If community complains about `revision` property, then indeed something is wrong.
Then it looks like this `revision` was introduced only for easier way to extract the project version during GHA release WF run.
Since there is a standard Maven way to get that info via `mvn help:evaluate -Dexpression="project.version" -q -DforceStdout` command, there is no reason in extra convenient property.

I'm not familiar with Maven flatten plugin, but introducing it just for `revision` property mitigation feels like an overhead in such a single module project.

And if you talk about alignment and deviations, there is [the reusable workflows project](https://github.com/spring-io/spring-github-workflows) which we use successfully in many places.
It might not be perfect, but it does exactly the same for many CI/CD processes and more.
I will be more than happy to migrate Spring Retry there as well. So, we would have one more project aligned and pointed only to a single place of truth or failure.

We may discuss this further offline.

snicoll:
>  If community complains about revision property, then indeed something is wrong.

Yeah, I am not denying that. I missed to add the last step in https://maven.apache.org/maven-ci-friendly.html, sorry about that.

> Since there is a standard Maven way to get that info via mvn help:evaluate -Dexpression="project.version" -q -DforceStdout command, there is no reason in extra convenient property.

I disagree with the analysis. The revision property has been introduced to mimic many other projects that use it. The link that I've shared above explains the benefit of using such property. The property hasn't been only introduced as a way to to get the version when releasing.

> And if you talk about alignment and deviations, there is [the reusable workflows project](https://github.com/spring-io/spring-github-workflows) which we use successfully in many places.

Sure, it's another reusable workflow you could use rather than this one, I don't have an opinion about that. But that's two different topic. Right now, it's neither.

artembilan:
Thank you for the link, Stephane!
Now I know more about Maven 😄 

So, I'll revert to `revision` and will look into this:

[...]

PR #481, opened 2024-12-18
GH-480: Bring back POM `revision` property

artembilan:
Fixes: https://github.com/spring-projects/spring-retry/issues/480

* Add `flatten-maven-plugin` to resolve properties and remove unnecessary build info from the final POM of the artifact to install/deploy

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>org.springframework.retry</groupId>
 	<artifactId>spring-retry</artifactId>
-	<version>2.0.12-SNAPSHOT</version>
+	<version>${revision}</version>
 	<name>Spring Retry</name>
 	<description><![CDATA[
 		Spring Retry provides an abstraction around retrying failed operations, with an
@@ -26,6 +26,7 @@
 		</license>
 	</licenses>
 	<properties>
+		<revision>2.0.12-SNAPSHOT</revision>
 		<disable.checks>false</disable.checks>
 		<java.version>17</java.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
@@ -298,6 +299,31 @@
 						<phase>package</phase>
 						<goals>
 							<goal>single</goal>
+						</goals>
+					</execution>
+				</executions>
+			</plugin>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>flatten-maven-plugin</artifactId>
+				<version>1.6.0</version>
+				<configuration>
+					<outputDirectory>${project.build.directory}</outputDirectory>
+					<flattenMode>oss</flattenMode>
+				</configuration>
+				<executions>
+					<execution>
+						<id>flatten</id>
+						<phase>process-resources</phase>
+						<goals>
+							<goal>flatten</goal>
+						</goals>
+					</execution>
+					<execution>
+						<id>flatten.clean</id>
+						<phase>clean</phase>
+						<goals>
+							<goal>clean</goal>
 						</goals>
 					</execution>
 				</executions>
--- a/pom.xml
+++ b/pom.xml
@@ -310,6 +310,9 @@
 				<configuration>
 					<outputDirectory>${project.build.directory}</outputDirectory>
 					<flattenMode>oss</flattenMode>
+					<pomElements>
+						<profiles>remove</profiles>
+					</pomElements>
 				</configuration>
 				<executions>
 					<execution>
@@ -317,13 +320,6 @@
 						<phase>process-resources</phase>
 						<goals>
 							<goal>flatten</goal>
-						</goals>
-					</execution>
-					<execution>
-						<id>flatten.clean</id>
-						<phase>clean</phase>
-						<goals>
-							<goal>clean</goal>
 						</goals>
 					</execution>
 				</executions>
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -39,6 +39,6 @@ runs:
       id: read-version
       shell: bash
       run: |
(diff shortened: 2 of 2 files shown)


artembilan:
The final result is as simple as this:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.springframework.retry</groupId>
  <artifactId>spring-retry</artifactId>
  <version>2.0.12-SNAPSHOT</version>
  <name>Spring Retry</name>
  <description>Spring Retry provides an abstraction around retrying failed operations, with an
		emphasis on declarative control of the process and policy-based behaviour that is
		easy to extend and customize. For instance, you can configure a plain POJO
		operation to retry if it fails, based on the type of exception, and with a fixed
		or exponential backoff.</description>
  <url>https://github.com/spring-projects/spring-retry</url>
  <organization>
    <name>Spring IO</name>
    <url>https://spring.io</url>
  </organization>
  <licenses>
    <license>
      <name>Apache 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>dsyer</id>
      <name>Dave Syer</name>
      <email>david.syer@broadcom.com</email>
    </developer>
    <developer>
      <id>garyrussell</id>
      <name>Gary Russell</name>
      <email>github@gprussell.net</email>
    </developer>
    <developer>
      <id>artembilan</id>
      <name>Artem Bilan</name>
      <email>artem.bilan@broadcom.com</email>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:git://github.com/spring-projects/spring-retry.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/spring-projects/spring-retry.git</developerConnection>
    <url>https://github.com/spring-projects/spring-retry</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>org.springframework</groupId>
      <artifactId>spring-context</artifactId>
      <version>6.0.23</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
    <dependency>
      <groupId>io.micrometer</groupId>
      <artifactId>micrometer-core</artifactId>
      <version>1.10.13</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
    <dependency>
      <groupId>com.github.spotbugs</groupId>
      <artifactId>spotbugs-annotations</artifactId>
      <version>4.8.6</version>
      <scope>compile</scope>
      <optional>true</optional>
    </dependency>
  </dependencies>
  <profiles>
    <profile>
      <id>spring</id>
      <repositories>
        <repository>
          <snapshots>
            <enabled>true</enabled>
          </snapshots>
          <id>spring-snapshots</id>
          <name>Spring Snapshots</name>
          <url>https://repo.spring.io/snapshot</url>
        </repository>
        <repository>
          <snapshots>
            <enabled>false</enabled>
          </snapshots>
          <id>spring-milestones</id>
          <name>Spring Milestones</name>
          <url>https://repo.spring.io/milestone</url>
        </repository>
      </repositories>
    </profile>
  </profiles>
</project>
```

Pay attention to the `<flattenMode>oss</flattenMode>` of the plugin.

snicoll (changes requested):
How about reverting the change in the release pipeline as well?

It should also add `.flattened-pom.xml` in `.gitignore`.

pom.xml
```diff
@@ -302,6 +303,31 @@
...
+						<phase>process-resources</phase>
+						<goals>
+							<goal>flatten</goal>
+						</goals>
+					</execution>
+					<execution>
+						<id>flatten.clean</id>
+						<phase>clean</phase>
+						<goals>
+							<goal>clean</goal>
```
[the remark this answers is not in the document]
artembilan: Not sure what you mean, but indeed don’t need `clean` here at all since I made the flatten result be in the `target` dir, which is cleaned by standard goal. Plus that would make no need for the gitignore as well.

artembilan:
Hey, Stephan!
Thank you for feedback!
I’ll revert GHA workflow, but IMHO what I did there is more Maven natural and it works with `revision` and without.
There is also similar commas to set version:
```
mvn versions:set -DnewVersion=2.0.12 -N versions:update-child-modules -DgenerateBackupPoms=false

**Review by snicoll — CHANGES_REQUESTED** [review]
`.flattened-pom.xml` is still missing in `.gitignore`.

File: pom.xml [hunk 2]
```diff
@@ -302,6 +303,27 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
+				<groupId>org.codehaus.mojo</groupId>
+				<artifactId>flatten-maven-plugin</artifactId>
+				<version>1.6.0</version>
+				<configuration>
+					<outputDirectory>${project.build.directory}</outputDirectory>
```
artembilan, line 311: > .flattened-pom.xml is still missing in .gitignore.

We don't need that because the file is stored into a `target` directory which is already under ignore.
I don't see any argument why this file has to be in the root project dir when we are interested in it only at runtime on `install` or `deploy`, so that is indeed a volatile entity for project build lifecycle and that is totally enough to have it in the `target` alongside with the rest of artifacts to deliver.

snicoll: If you do that, you don't need `clean` then. Again, if you decide to do things differently, please do that consistently.

artembilan: Right. That’s what I did in the last commit. Please, double check the latest change set in the PR

snicoll (approved):

PR #490, opened 2025-05-07
Fix proxy attribute name mismatch in RetryOperationsInterceptor

junhyeongkim2:
* Changed removeAttribute to use the same key as setAttribute("___proxy___").
* Fixed an issue where a proxy reference remained in the context due to the previous "__proxy__" key mismatch. 

Found a small bug: I might be mistaken, but I think the proxy key mismatch is preventing removal. Please have your team take a look when you have a moment. Thanks!

### DIFF
--- a/src/test/java/org/springframework/retry/interceptor/RetryOperationsInterceptorTests.java
+++ b/src/test/java/org/springframework/retry/interceptor/RetryOperationsInterceptorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2006-2023 the original author or authors.
+ * Copyright 2006-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -49,6 +49,8 @@ import org.springframework.util.ClassUtils;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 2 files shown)


fengkongling:
ok，亲~~~

gaofengIt:
您好，我已经收到您的邮件，将尽快给您回复。

artembilan (approved):

artembilan (changes requested):
Please, consider to use your legal name for Git client.
That `Signed-off-by: junhyeongkim2 <ggprgrkjh@naver.com>` is not a official name and we cannot accept such a contribution.

See more about DCO: https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin

junhyeongkim2:
I've updated the DCO signature with my legal name as requested. Thank you for your guidance!

junhyeongkim2:
I've removed the duplicate DCO signatures and kept only my legal name version. The issue is now resolved.

artembilan (changes requested):
I wonder if that would be possible to cover your fix with some unit test.
Apparently we don't have one since this was wrong there for years.
Thanks

src/test/java/org/springframework/retry/interceptor/RetryOperationsInterceptorTests.java
```diff
@@ -257,6 +260,70 @@ public Object proceed() {
...
+	}
+
+	@Test
+	public void testProxyAttributeCleanupEvenWhenIllegalStateExceptionThrown() {
+		RetryContext context = new RetryContextSupport(null);
+		Object mockProxy = new Object();
+		RetrySynchronizationManager.register(context);
+		context.setAttribute("___proxy___", mockProxy);
+		assertThat(context.getAttribute("___proxy___")).isNotNull();
+		assertThatIllegalStateException().isThrownBy(() -> this.interceptor.invoke(new MethodInvocation() {
```
artembilan, line 296: Please, consider to extract a `MethodInvocation` into a variable before this assertion.
After that the assertion sentence would be much easier to read.
I also don't see a reason to have the whole interface implementation.
We probably can simply use a `mock()` and then stub that `getMethod()`.

artembilan (changes requested):
Please, also update Copyright of the affected classes to current `2025`.
Thanks

src/test/java/org/springframework/retry/interceptor/RetryOperationsInterceptorTests.java
junhyeongkim2: Thank you for your feedback! I’ve applied all of your suggestions. Please let me know if you’d like any further adjustments.

artembilan:
@junhyeongkim2 ,

thank you very much for the contribution; looking forward for more!

PR #492, opened 2025-05-11
Add warnings for invalid setter values in ExponentialBackoffPolicy

### LINKED ISSUE #352: Inconsistent validation of exponential backoff
[setMultiplier()](https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java#L140) in ExponentialBackoffPolicy allows a minimum multiplier of 1.0. However, the [corresponding builder method in RestTemplateBuilder](https://github.com/spring-projects/spring-retry/blob/main/src/main/java/org/springframework/retry/support/RetryTemplateBuilder.java#L202) insists on a minimum multiplier greater than 1.0. There's even a test for it.

That's confusing. While it makes sense to require that the multiplier is greater than 1.0, aligning the setter with the builder might break existing usages.

kssumin:
This PR addresses #352 by adding warning logs when invalid values are provided to ExponentialBackoffPolicy setters.

## Changes
- Added warning logs in `setInitialInterval()` when value < 1
- Added warning logs in `setMultiplier()` when value <= 1.0
- Added warning logs in `setMaxInterval()` when value <= 0

## Motivation
The current implementation silently adjusts invalid values without informing users. This can lead to confusion when configured values differ from actual behavior. Adding warning logs helps users understand when their configuration needs adjustment.

## Backward Compatibility
This change maintains backward compatibility by:
- Keeping the existing behavior of silently adjusting invalid values
- Only adding warning logs to inform users
- No breaking changes to the API

## Related Issue
Closes #352

### DIFF
--- a/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
+++ b/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
@@ -43,6 +43,7 @@ import org.springframework.util.ClassUtils;
  * @author Artem Bilan
  * @author Marius Lichtblau
  * @author Anton Aharkau
+ * @author Kim Sumin
  */
 @SuppressWarnings("serial")
 public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<ExponentialBackOffPolicy> {
@@ -130,6 +131,9 @@ public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<Exponenti
 	 * @param initialInterval the initial interval
 	 */
 	public void setInitialInterval(long initialInterval) {
+		if (initialInterval < 1) {
+			logger.warn("Initial interval must be at least 1, but was " + initialInterval);
+		}
 		this.initialInterval = initialInterval > 1 ? initialInterval : 1;
 	}
 
@@ -139,6 +143,9 @@ public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<Exponenti
 	 * @param multiplier the multiplier
 	 */
 	public void setMultiplier(double multiplier) {
+		if (multiplier <= 1.0) {
+			logger.warn("Multiplier must be > 1.0 for effective exponential backoff, but was " + multiplier);
+		}
 		this.multiplier = multiplier > 1.0 ? multiplier : 1.0;
 	}
 
@@ -150,6 +157,9 @@ public class ExponentialBackOffPolicy implements SleepingBackOffPolicy<Exponenti
 	 * @param maxInterval in milliseconds.
 	 */
 	public void setMaxInterval(long maxInterval) {
+		if (maxInterval < 1) {
+			logger.warn("Max interval must be positive, but was " + maxInterval);
+		}
 		this.maxInterval = maxInterval > 0 ? maxInterval : 1;
 	}
 
--- a/src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
+++ b/src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
@@ -16,10 +16,15 @@
 
 package org.springframework.retry.backoff;
 
+import org.apache.commons.logging.Log;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
+import org.springframework.beans.DirectFieldAccessor;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 
 /**
  * @author Rob Harrop
@@ -27,6 +32,7 @@ import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
  * @author Gary Russell
  * @author Marius Lichtblau
  * @author Anton Aharkau
+ * @author Kim Sumin
  */
 public class ExponentialBackOffPolicyTests {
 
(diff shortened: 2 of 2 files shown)


src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -252,6 +269,18 @@ public void backOff(BackOffContext backOffContext) throws BackOffInterruptedExce
...
 
+	/**
+	 * Setter for {@link Log}. If not applied the following is used:
+	 * <p>
+	 * {@code LogFactory.getLog(getClass())}
+	 * </p>
+	 * @param logger the logger the exponential backoff policy uses for logging
+	 * @since 2.0.x
+	 */
+	public void setLogger(Log logger) {
```
artembilan, line 280: This is not correct. Such a functionality must not be exposed to end-user.
Please, consider to use a reflection in your tests instead.
See `DirectFieldAccessor`, for example.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -130,6 +131,11 @@ protected void cloneValues(ExponentialBackOffPolicy target) {
 	 * @param initialInterval the initial interval
 	 */
 	public void setInitialInterval(long initialInterval) {
+		if (initialInterval < 1) {
+			logger.warn("The initial interval value " + initialInterval +
+					" is less than 1. " +
+					"The initial interval should be at least 1. ");
```
artembilan, line 137: Both sentences say exactly the same. 
Why cannot we use just one?

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -139,6 +145,12 @@ public void setInitialInterval(long initialInterval) {
 	 * @param multiplier the multiplier
 	 */
 	public void setMultiplier(double multiplier) {
+		if (multiplier <= 1.0) {
+			logger.warn("The multiplier value " + multiplier +
+					" is less than or equal to 1.0. " +
+					"For exponential backoff to work effectively, " +
+					"the multiplier should be greater than 1.0. ");
```
artembilan, line 152: Same here.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
```diff
@@ -150,6 +162,11 @@ public void setMultiplier(double multiplier) {
 	 * @param maxInterval in milliseconds.
 	 */
 	public void setMaxInterval(long maxInterval) {
+		if (maxInterval < 1) {
+			logger.warn("The max interval value " + maxInterval +
+					" is less than or equal to 0. " +
+					"The max interval should be positive.");
```
artembilan, line 168: and here.

src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
```diff
@@ -16,17 +16,21 @@
 
 package org.springframework.retry.backoff;
 
+import org.apache.commons.logging.Log;
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentCaptor;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.*;
```
artembilan, line 25: No asterisk imports, even in tests.

artembilan (changes requested):

artembilan (changes requested):
```
Error:  Failed to execute goal io.spring.javaformat:spring-javaformat-maven-plugin:0.0.43:validate (default) on project spring-retry: Formatting violations found in the following files:
Error:   * /home/runner/work/spring-retry/spring-retry/src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
Error:   * /home/runner/work/spring-retry/spring-retry/src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
Error:  
Error:  Run `spring-javaformat:apply` to fix.
```

kssumin:
@artembilan Thank you for the detailed review! I've addressed all your feedback:

1. Removed the `setLogger()` method as it shouldn't be exposed to end-users
2. Updated all test methods to use `DirectFieldAccessor` for reflection-based logger access
3. Simplified all warning messages to avoid redundancy:
   - Initial interval: "Initial interval must be at least 1, but was X"
   - Multiplier: "Multiplier must be > 1.0 for effective exponential backoff, but was X"
   - Max interval: "Max interval must be positive, but was X"
4. Applied spring-javaformat to fix formatting violations

All changes have been implemented in the latest commit. Please let me know if there's anything else that needs to be adjusted.

src/main/java/org/springframework/retry/backoff/ExponentialBackOffPolicy.java
kssumin: @artembilan Thanks for the insightful feedback about DirectFieldAccessor! I learned a lot about the importance of not polluting the public API with test-specific methods. Really appreciate the learning opportunity!

artembilan (changes requested):
Looks good!
Please, add your name to the `@author` list of the effected classes.

artembilan (changes requested):
```
Error:   * /home/runner/work/spring-retry/spring-retry/src/test/java/org/springframework/retry/backoff/ExponentialBackOffPolicyTests.java
Error:  
Error:  Run `spring-javaformat:apply` to fix.
```
Please, ensure to run `mvnw verify` before pushing to the PR.
Thanks

kssumin:
스크린샷 2025-05-16 오전 2 19 30

@artembilan 
Thank you for the review! I've run spring-javaformat:apply and completed the verification. The build was successful as shown in the screenshot. This is my first time contributing to open source, so I wasn't familiar with these formatting rules. I appreciate your guidance! I'll make sure to run 'mvnw verify' before pushing to PRs in the future. 😊

artembilan (changes requested):
Last remark.
I see this in the commit:
```
Signed-off-by: kssumin <201566@jnu.ac.kr>
```
That has to be your real legal name.
We cannot accept contributions with nick names.
Also, see if that is real email.

The DCO requirements must be met.

You have to fix your Git client to use your proper legal name.
Then you would need to force-push into this PR with new `Signed-off-by`.

Thank you for understanding!

kssumin:
The email was indeed valid, but I've now changed it to a more commonly used email address.
Thank you for explaining these requirements in detail. I appreciate your guidance and have learned a lot from this process.

artembilan (approved):
Cool! Will merge tomorrow , my morning.
Thank you again!

PR #493, opened 2025-06-24
only use specified listeners in `@Retryable`

### LINKED ISSUE #485: Add a possibility to use listener per @Retryable method instead of global listeners
In my project I wanted to add a listener and use it along with `@Retryable` annotation.
But when I add a listener, it's automatically considered by Spring Retry as global one, meaning it will be applied at least to every method marked with `@Retryable` annotation, which prevents me from creating listener, because I don't want to impact an existing code.

Example:
`@Retryable(retryFor = SomeException.class, maxAttempts = 5, listeners = "myListenerBean")`

When I create myListenerBean, it will be automatically applied to all existing methods with `@Retryable` annotation, which is not desired behavior.
The only way to prevent this is to either specify an empty string on existing code, which is not always possible **to modify an existing code**

`@Retryable(listeners = "")`

or use interceptor attribute which cannot be used with any other attributes and which requires interceptor bean along with all retry configurations code:

`@Retryable(interceptor = "myInterceptorBean")`

As a result instead of using neat single line:
`@Retryable(retryFor = SomeException.class, maxAttempts = 5, listeners = "myListenerBean")`

I'm forced to use:
`@Retryable(interceptor = "myInterceptorBean")`

along with whole retry configuration, for example:
```
@Slf4j
@Configuration
@RequiredArgsConstructor
public class ConnectionErrorRetryConfig {
    public static final String CONNECTION_ERROR_RETRY_INTERCEPTOR = "connectionErrorRetryInterceptor";

    @Value("${request.max-attempts:5}")
    private int maxAttempts;

    private final MyService myService;

    @Bean(CONNECTION_ERROR_RETRY_INTERCEPTOR)
    public RetryOperationsInterceptor connectionErrorRetryInterceptor() {
        SimpleRetryPolicy simpleRetryPolicy = new SimpleRetryPolicy();
        simpleRetryPolicy.setMaxAttempts(maxAttempts);

        RetryTemplate retryTemplate = RetryTemplate.builder()
            .customPolicy(simpleRetryPolicy)
            .retryOn(SomeException.class)
            // forced to create a new RetryListener here as NOT a bean, because bean is impacting globally every @Retryable method
            .withListener(new RetryListener() {
                @Override
                public <T, E extends Throwable> void onError(RetryContext context, RetryCallback<T, E> callback, Throwable throwable)
                {
                    if (throwable instanceof SomeException someException)
                    {
                        myService.doImportantAction(someException);
                    }

                    RetryListener.super.onError(context, callback, throwable);
                }
            })
            .build();

        return RetryInterceptorBuilder.stateless()
            .retryOperations(retryTemplate)
            .recoverer(new ConnectionErrorRecoverer())
            .build();
    }
}
```

As it's been already discussed [here](https://github.com/spring-projects/spring-retry/discussions/484#discussion-7916466), we could think of any alternative, for example deprecating existing listeners attribute and/or providing some other attribute(s) etc. in order to not use global listeners.

Thanks.

mjd507:
Fixes: #485

### DIFF
--- a/README.md
+++ b/README.md
@@ -464,12 +464,15 @@ public class Application {
 
 @Service
 class Service {
-    @Retryable(RemoteAccessException.class)
+    @Retryable(retryFor = RemoteAccessException.class, listeners = { "retryListener1", "retryListener2" })
     public service() {
         // ... do something
     }
 }
 ```
+
+Starting with version 2.0.13, any bean instance of `RetryListener` are not considered as global listeners.  
+You always have to explicitly specify the `listeners` attribute in `@Retryable` if needed.
 
 You can use the attributes of `@Retryable` to control the `RetryPolicy` and `BackoffPolicy`, as follows:
 
--- a/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
+++ b/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java
@@ -21,6 +21,7 @@ import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -77,6 +78,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Roman Akentev
  * @author Aftab Shaikh
+ * @author Jiandong Ma
  * @since 1.1
  */
 public class AnnotationAwareRetryOperationsInterceptor implements IntroductionInterceptor, BeanFactoryAware {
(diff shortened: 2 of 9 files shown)


gaofengIt:
您好，我已经收到您的邮件，将尽快给您回复。

fengkongling:
ok，亲~~~

mjd507:
@artembilan kindly review and suggest.
Thank you

README.md
```diff
@@ -464,13 +464,16 @@ public class Application {
...
-    @Retryable(RemoteAccessException.class)
+    @Retryable(retryFor = RemoteAccessException.class, listeners = { "retryListener1", "retryListener2" })
     public service() {
         // ... do something
     }
 }
 ```
 
+Starting with version 2.0.13, any bean instance of `RetryListener` are not considered as global listeners.  
+You always have to explicitly specify the `listeners` attribute in `@Retryable` if needed.
```
artembilan, line 475: And this is exactly why it has not been fixed yet.
This is a breaking change which we don't do according to our backward compatibility support.
This makes sense for a major version and we are not there yet.
We can keep your PR for time being though!

artembilan (changes requested):
@mjd507 ,

thank you for contribution, but we cannot accept it at the moment as we follow no breaking change in the point version.

README.md
mjd507: got it. thank you

artembilan:

Issue #496, opened 2025-10-01
[Bug] RecoverAnnotationRecoveryHandler chooses wrong @Recover method when JVM returns non-deterministic getDeclaredMethods order


chihyu0917:
### Summary

`RecoverAnnotationRecoveryHandler` occasionally invokes an *unexpected* `@Recover`
method when more than one candidate is a valid match.
The outcome depends on the **order of `Class#getDeclaredMethods()`**, which is not
specified by the JVM.  
In real builds this shows up as flaky tests or different behaviour between
JDK distributions/OS-es.

### Current behaviour

* If several `@Recover` methods are *equally* suitable (same distance in the
  exception hierarchy, same parameter compatibility), the handler iterates over
  `this.methods.entrySet()` — a `HashMap` backed by whatever order the JVM gave
  us — and returns the *last* method that satisfies the tie-break rules.
* Because that order changes, the selected method is not deterministic.

### Expected behaviour

Given the same set of `@Recover` methods the handler should **always** choose the
same target method, independent of JVM or library versions.

### Reproduction

The built-in tests already expose the problem when run under
[NonDex 2.1.7](https://github.com/TestingResearchIllinois/NonDex):

| test | command (20 random seeds) | typical failure |
|------|---------------------------|-----------------|
| *extends Throwable* | `mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.springframework.retry.annotation.RecoverAnnotationRecoveryHandlerTests#multipleQualifyingRecoverMethodsExtendsThrowable -DnondexRuns=20` | `expected: 2 but was: 1` |
| *no Throwable* | `mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.springframework.retry.annotation.RecoverAnnotationRecoveryHandlerTests#multipleQualifyingRecoverMethodsWithNoThrowable -DnondexRuns=20` | `IllegalArgumentException` |
| *null Throwable* | `mvn -pl . edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=org.springframework.retry.annotation.RecoverAnnotationRecoveryHandlerTests#multipleQualifyingRecoverMethodsWithNull -DnondexRuns=20` | `IllegalArgumentException` |

### Analysis

* `HashMap` order + JVM reflection order ⇒ non-deterministic.
* When two candidates tie, the current algorithm does **not** consider
[...]

PR #497, opened 2025-10-01
fix: make RecoverAnnotationRecoveryHandler deterministic

chihyu0917:
This PR supersedes discussion in #496 
When multiple @Recover candidates exist, choose the most specific and deterministic handler instead of relying on reflection order. Previously the selected @Recover method depended on HashMap iteration order. Switching to LinkedHashMap + deterministic sorting removes this problem. Verified by NonDex.

### DIFF
--- a/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
+++ b/src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
@@ -19,8 +19,13 @@ package org.springframework.retry.annotation;
 import java.lang.reflect.Method;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.HashMap;
+import java.util.Arrays;
+import java.util.List;
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.springframework.classify.SubclassClassifier;
 import org.springframework.core.annotation.AnnotatedElementUtils;
@@ -55,12 +60,13 @@ import org.springframework.util.StringUtils;
  * @author Gianluca Medici
  * @author Lijinliang
  * @author Yanming Zhou
+ * @author Chih-Yu Huang
  */
 public class RecoverAnnotationRecoveryHandler<T> implements MethodInvocationRecoverer<T> {
 
 	private final SubclassClassifier<Throwable, Method> classifier = new SubclassClassifier<>();
 
-	private final Map<Method, SimpleMetadata> methods = new HashMap<>();
+	private final Map<Method, SimpleMetadata> methods = new LinkedHashMap<>();
 
 	private final Object target;
 
@@ -73,7 +79,8 @@ public class RecoverAnnotationRecoveryHandler<T> implements MethodInvocationReco
 
 	@Override
 	public T recover(Object[] args, Throwable cause) {
-		Method method = findClosestMatch(args, cause.getClass());
+		Class<? extends Throwable> causeType = (cause == null) ? null : cause.getClass();
+		Method method = findClosestMatch(args, causeType);
 		if (method == null) {
 			throw new ExhaustedRetryException("Cannot locate recovery method", cause);
 		}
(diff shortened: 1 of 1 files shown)


src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -330,4 +357,4 @@ public Object[] getArgs(Throwable t, Object[] args) {
 
 	}
 
-}
+}
```
artembilan: Please, ensure a new line in the end of file.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -73,6 +79,7 @@ public RecoverAnnotationRecoveryHandler(Object target, Method method) {
 
 	@Override
 	public T recover(Object[] args, Throwable cause) {
+		Class<? extends Throwable> causeType = (cause == null) ? null : cause.getClass();
 		Method method = findClosestMatch(args, cause.getClass());
```
artembilan: You have just extracted the type, but don't use it here.

artembilan (changes requested):
```
Signed-off-by: chihyuh3 <chihyuh3@fa25-cs527-010.cs.illinois.edu>
```
Please, use your legal name according to DCO requirements.
You can configure your Git client to provide a proper name.

Please, also add your name to the `@author` list of the affected class.

chihyu0917:
Thanks for your review! I corrected those mistakes that you mentioned.
* [x] Commit(s) contain `Signed-off-by:` trailer (DCO) with legal name
* [x] Add my name to the `@author` list 
* [x] End-of-File 
* [x] Now use `causeType` variable

artembilan (changes requested):
Please, run tests locally (`./mvnw clean verify`) before pushing:
```
Error:  Failures: 
Error:    EnableRetryNoThreadLocalTests>EnableRetryTests.recoveryWithoutParam:151 
expected: "test"
 but was: null
Error:    EnableRetryTests.recoveryWithoutParam:151 
expected: "test"
 but was: null
Error:    RecoverAnnotationRecoveryHandlerTests.noMatch:105 
Expecting actual throwable to be an instance of:
  org.springframework.retry.ExhaustedRetryException
but was:
  java.lang.IllegalArgumentException: argument type mismatch
```

Also, pay attention to Spring Java Format plugin: https://github.com/spring-io/spring-javaformat.

chihyu0917:
Thanks for the review and feedback! My apologies for the regressions in the previous attempt.

I have now run the full `./mvnw clean verify` to ensure all tests pass and have applied the spring-javaformat plugin for code style.

The root cause of the non-determinism was twofold: the unpredictable order of methods returned by reflection, and the use of unordered HashMaps which destroyed any stable ordering.

This PR now makes the entire selection process deterministic by:

Sorting Discovered Methods: The init method now sorts all found @Recover methods alphabetically, ensuring a stable initial processing order.
Preserving Order: All internal Maps used for storing and filtering these methods have been changed to LinkedHashMap to preserve this stable order throughout the handler's lifecycle.
Deterministic Tie-Breaking: The selection logic now first finds all candidates with the best exception-distance, and then deterministically picks the most specific one (by parameter count) from that stable, sorted list.

This approach resolves the original issue and passes all existing tests. Ready for another look. Thanks!

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -112,50 +119,91 @@ private Method findMethodOnProxy(Method method, Object proxy) {
...
+		}
+
+		Method result = findMethodWithThrowable(args, cause, withThrowable);
+		if (result == null) {
+			result = findMethodWithNoThrowable(args, withoutThrowable);
+		}
+		return result;
+	}
+
+	private Method findMethodWithNoThrowable(Object[] args, List<Method> methods) {
```
artembilan: Looks like this can be `static` and therefore goes to the end of class.

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -112,50 +119,91 @@ private Method findMethodOnProxy(Method method, Object proxy) {
...
+					if (compareParameters(args, method.getParameterTypes(), exceptionType != null)) {
+						return method;
+					}
+				}
+			}
+		}
+		return null;
+	}
+
+	private int calculateDistance(Class<?> cause, Class<?> type) {
```
artembilan: Also `static`

src/main/java/org/springframework/retry/annotation/RecoverAnnotationRecoveryHandler.java
```diff
@@ -165,53 +213,59 @@ private int calculateDistance(Class<? extends Throwable> cause, Class<? extends
 		return result;
 	}
 
-	private boolean compareParameters(Object[] args, int argCount, Class<?>[] parameterTypes,
-			boolean withRecoverMethodName) {
-		if ((withRecoverMethodName && argCount == args.length) || argCount == (args.length + 1)) {
-			int startingIndex = 0;
-			if (parameterTypes.length > 0 && Throwable.class.isAssignableFrom(parameterTypes[0])) {
-				startingIndex = 1;
+	private boolean compareParameters(Object[] args, Class<?>[] parameterTypes, boolean hasThrowable) {
```
artembilan: `static`?

artembilan (changes requested):

chihyu0917:
Following your suggestion, I’ve converted them to `static`. Thanks for providing your suggestions!

artembilan:
Thank you; and looking forward for more!

chihyu0917:
Great, thanks!

